PR #9, opened 2013-03-21
Adding support for NewPage

DavidGriffith:
Here's some groundwork for getting explicit page breaks into Pandoc


mb21:
I think this three-year old, barely functional pull request can safely be closed, see https://github.com/jgm/pandoc/issues/1934 for the ongoing discussion.

PR #21, opened 2016-05-02
Use `<$` instead of `fmap . const`

sid-kap:
Not sure if this will work with old Preludes but let's run the Travis tests and see...


sid-kap:
This won't work for base-4.0.0.0, so never mind.

PR #23, opened 2016-09-30
Reimplement ToJSON / FromJSON instances, and update format

jkr:
1. This adds manual ToJSON and FromJSON instances, instead of the generic-derivied instances we had previously used. This should keep the format stable, regardless of  changes in aeson. Empirically, it also adds a small speedup in encoding and decoding.

2. This updates the pandoc json format. There are three main changes:

  1. The toplevel format is now
      
       ~~~{.json}
      {
         "pandoc-api-version": [1,7,1],
         "meta": [meta*],
         "blocks": [blocks*]
      }
      ~~~

      where `meta` and `blocks` are individual `Meta` and `Block` elements. Note that this means that the   toplevel is now an object, and that `meta` elements are no longer under an `unMeta` key.

  2. The "pandoc-api-version" tracks the pandoc-types version. If the major and minor versions (the 0th and 1st elements of the array) do not match, decoding in pandoc will fail. It is up to library writers in other languages to check and write these versions.

  3. Leaf elements (elements with no contents, like `Space`, `LineBreak`, and `HorinzontalRule`) will no longer have a `"c"` element. I.e. instead of `{"t": "Space", "c": []}`, we will output `{"t": "Space"}`. This probably won't affect many decoding functions, but it will be necessary for library writers to update their encoding functions for these elements.

3. Finally, this adds Arbitrary instances to pandoc-types, in `Text.Pandoc.Arbitrary`.

### DIFF
--- a/Text/Pandoc/Definition.hs
+++ b/Text/Pandoc/Definition.hs
@@ -1,4 +1,7 @@
-{-# LANGUAGE DeriveDataTypeable, DeriveGeneric, FlexibleContexts, CPP #-}
+{-# LANGUAGE OverloadedStrings, OverloadedLists, DeriveDataTypeable,
+             DeriveGeneric, FlexibleContexts, GeneralizedNewtypeDeriving,
+             CPP
+#-}
 
 {-
 Copyright (c) 2006-2016, John MacFarlane
@@ -74,7 +77,7 @@ module Text.Pandoc.Definition ( Pandoc(..)
 
 import Data.Generics (Data, Typeable)
 import Data.Ord (comparing)
-import Data.Aeson (FromJSON(..), ToJSON(..))
+import Data.Aeson
 import qualified Data.Aeson.Types as Aeson
 import Control.Monad (guard)
 import qualified Data.Map as M
@@ -196,7 +199,7 @@ type TableCell = [Block]
 
 -- | Formats for raw blocks
 newtype Format = Format String
-               deriving (Read, Show, Typeable, Data, Generic)
+               deriving (Read, Show, Typeable, Data, Generic, ToJSON, FromJSON)
 
 instance IsString Format where
   fromString f = Format $ map toLower f
@@ -321,24 +324,50 @@ instance ToJSON Meta
 
 instance FromJSON CitationMode
   where parseJSON = parseJSON'
-instance ToJSON CitationMode
-  where toJSON = toJSON'
+instance ToJSON CitationMode where
+  toJSON cmode =
+    object [ "t" .= String s
+           , "c" .= Aeson.emptyArray
+           ]
+    where s = case cmode of
+            AuthorInText   -> "AuthorInText"
+            SuppressAuthor -> "SuppressAuthor"
+            NormalCitation -> "NormalCitation"
+
 
 instance FromJSON Citation
   where parseJSON = parseJSON'
-instance ToJSON Citation
-  where toJSON = toJSON'
+instance ToJSON Citation where
+  toJSON cit =
+    object [ "citationId"      .= citationId cit
+           , "citationPrefix"  .= citationPrefix cit
+           , "citationSuffix"  .= citationSuffix cit
+           , "citationMode"    .= citationMode cit
+           , "citationNoteNum" .= citationNoteNum cit
+           , "citationHash"    .= citationHash cit
+           ]
 
 instance FromJSON QuoteType
   where parseJSON = parseJSON'
-instance ToJSON QuoteType
-  where toJSON = toJSON'
+instance ToJSON QuoteType where
+  toJSON qtype = object [ "t" .= String s
+                        , "c" .= Aeson.emptyArray
+                        ]
+    where s = case qtype of
+            SingleQuote -> "SingleQuote"
+            DoubleQuote -> "DoubleQuote"
+
 
 instance FromJSON MathType
   where parseJSON = parseJSON'
-instance ToJSON MathType
-  where toJSON = toJSON'
-
+instance ToJSON MathType where
+  toJSON mtype = object [ "t" .= String s
+                        , "c" .= Aeson.emptyArray
+                        ]
+    where s = case mtype of
+            DisplayMath -> "DisplayMath"
+            InlineMath  -> "InlineMath"
+  
 instance FromJSON ListNumberStyle
   where parseJSON = parseJSON'
 instance ToJSON ListNumberStyle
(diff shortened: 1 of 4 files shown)


jkr:
Failures are because of OverloadedLists in 7.4, 7.6. We should probably update the travis.yaml to match pandoc's.

Text/Pandoc/Definition.hs
```diff
@@ -321,24 +324,50 @@ instance ToJSON Meta
 
 instance FromJSON CitationMode
   where parseJSON = parseJSON'
-instance ToJSON CitationMode
-  where toJSON = toJSON'
+instance ToJSON CitationMode where
+  toJSON cmode =
+    object [ "t" .= String s
```
jgm: I think it would be a good idea to abstract out this pattern, to avoid all the code repetition.   Also, is there a reason not to use Show instances?  Then this could be something like

    toJSON cmode = tagged (show cmode) Nothing

In cases where we have a nonempty content, we could have (e.g.)

    toJSON (Emph ils) = tagged "Emph" ils

This would make the code more compact and decrease the possibility of small random errors.

jgm:
This looks great!  I made one comment about some easy things to reduce code duplication.
I think changing travis.yaml to bring it in line with the versions we support in pandoc makes good sense.

jkr:
Good point. It has to be two separate functions because the rigid type signature in `toJSON a =>`  causes problems if we try to put a Maybe around it (even if it's a Nothing, aeson won't skip it unless it knows what *kind* of Nothing it is). But keeping two functions in sync should be easier.

I also switched out the explicit encoding of arrays with making use of tuples. This is a bit fragile, because instances are only encoded up to 7-tuples, but the highest we have is tables with, I think, 5. In general, encoding (heterogeneous) arrays is a bit annoying in Aeson because you can't have heterogeneous lists.
This means we don't need OverloadedLists, so we'll pass tests for earlier compilers, for whatever it's worth.

Later on, today or tomorrow, I'll try to put in a proper test framework with some hunit tests -- we test encode/decode round trip, but that doesn't guarantee we're actually writing or reading the correct format.

jgm:
+++ Jesse Rosenthal [Oct 01 16 07:42 ]:
>   Good point. It has to be two separate functions because the rigid type
>   signature in toJSON a => causes problems if we try to put a Maybe
>   around it (even if it's a Nothing, aeson won't skip it unless it knows
>   what kind of Nothing it is). But keeping two functions in sync should
>   be easier.

What I had in mind was just one function,

    tagged :: ToJSON a => [Char] -> Maybe a -> Value
    tagged x (Just y) = object [ "t" .= x, "c" .= y ]
    tagged x Nothing  = object [ "t" .= x ]

But your approach is less typing overall, so it's probably
better.

>   I also switched out the explicit encoding of arrays with making use of
>   tuples. This is a bit fragile, because instances are only encoded up to
>   7-tuples, but the highest we have is tables with, I think, 5. In
>   general, encoding (heterogeneous) arrays is a bit annoying in Aeson
>   because you can't have heterogeneous lists.

Okay.  In JSON they're both coming across as lists, right?
You can sort of have heterogeneous lists in JSON, but you
have to make them lists of Values, so this is more
cumbersome.

jkr:
Right, but that has to be explicitly typed Inline  every time you use the Nothing, or else the compiler complains it can't resolve the `a`. And there isn't a type, since there isn't a value, so you have to say `Nothing :: Maybe ()` each time (or any type -- it doesn't matter). Which makes the function really annoying to use.
On Oct 1, 2016 2:35 PM, John MacFarlane <notifications@github.com> wrote:+++ Jesse Rosenthal [Oct 01 16 07:42 ]:
>   Good point. It has to be two separate functions because the rigid type
>   signature in toJSON a => causes problems if we try to put a Maybe
>   around it (even if it's a Nothing, aeson won't skip it unless it knows
>   what kind of Nothing it is). But keeping two functions in sync should
>   be easier.

What I had in mind was just one function,

    tagged :: ToJSON a => [Char] -> Maybe a -> Value
    tagged x (Just y) = object [ "t" .= x, "c" .= y ]
    tagged x Nothing  = object [ "t" .= x ]

But your approach is less typing overall, so it's probably
better.

>   I also switched out the explicit encoding of arrays with making use of
>   tuples. This is a bit fragile, because instances are only encoded up to
>   7-tuples, but the highest we have is tables with, I think, 5. In
>   general, encoding (heterogeneous) arrays is a bit annoying in Aeson
>   because you can't have heterogeneous lists.

Okay.  In JSON they're both coming across as lists, right?
You can sort of have heterogeneous lists in JSON, but you
have to make them lists of Values, so this is more
cumbersome.

jgm:
+++ Jesse Rosenthal [Oct 01 16 11:42 ]:
>   Right, but that has to be explicitly typed Inline  every time you use
>   the Nothing, or else the compiler complains it can't resolve the `a`.
>   And there isn't a type, since there isn't a value, so you have to say
>   `Nothing :: Maybe ()` each time (or any type -- it doesn't matter).
>   Which makes the function really annoying to use.

OK, makes sense now.

jkr:
> Okay. In JSON they're both coming across as lists, right?
> You can sort of have heterogeneous lists in JSON, but you
> have to make them lists of Values, so this is more
> cumbersome.
The output is the same, and correct. This is a shortcut. But if you have long heterogeneous lists you have to take the long way around.

jgm:
@jkr are you still planning to add some unit tests?
Also, this is a chance for us to make other breaking changes in pandoc-types.  But I don't want to delay a pandoc release too long, so I don't want to make changes that are going to require heavy changes in pandoc unless we can get them done quickly.

jkr:
I'm slowly adding the units: see 

https://github.com/jkr/pandoc-types/blob/testJSON/test/test-pandoc-types.hs 

I'm through most of the Inlines, so more than halfway done, but it's pretty slow and boring work, so I've only been doing a few a day.

I don't see any more breaking changes that need to be made, personally.

tarleb:
> Also, this is a chance for us to make other breaking changes in pandoc-types. But I don't want to delay a pandoc release too long, so I don't want to make changes that are going to require heavy changes in pandoc unless we can get them done quickly.

How about the lineblock element (#22), do you think it could be included? What would have to be changed to move forward with this?

jgm:
+++ Albert Krewinkel [Oct 09 16 03:04 ]:
>     Also, this is a chance for us to make other breaking changes in
>     pandoc-types. But I don't want to delay a pandoc release too long,
>     so I don't want to make changes that are going to require heavy
>     changes in pandoc unless we can get them done quickly.
>
>   How about the lineblock element ([1]#22), do you think it could be
>   included? What would have to be changed to move forward with this?

Sorry, I haven't had a chance to look through the associated
pandoc changes.  Have you modified all the readers and
writers to handle the new element?

tarleb:
All writers have been modified, using lineblock, verse, or `pre`-like environments for the new element in AsciiDoc, ConTeXt, HTML, Markdown, Org, and RST.  For all other output formats, the element is treated like a `Para` with hard line breaks.

I thought about using a `verse` environment in the LaTeX writer, it's what Org-mode uses to export a `verse` block.  However, that seemed overly heavy and is probably oftentimes not what people would expect, so I decided against it.  I can add it back, if you'd prefer it over the para/hardbreak approach there.

The Docbook5, Markdown, Org, and RST readers can use lineblock.  The LaTeX reader currently doesn't use the LineBreak element: for symmetry with the writer, but also because I wasn't sure how to deal with multi-stanza poems and linebreaks in emphasis (`\emph{one\\ two}`).

jgm:
OK, I'm going to merge the pandoc-types changes.
I've made a few requests about rebasing the pandoc changes.

@tarleb can you add a unit test for line blocks to
test/test-pandoc-types.hs?

PR #24, opened 2016-10-06
Add explicit unit tests for encoding and decoding.

jkr:
Previously we had just done round-trip tests, but that can hide a
mistake if the encoder and decoder are making the same mistake (or if
the encoder makes a mistake that the decoder ignores). This tests each
pandoc entity against a specific JSON string. There is some redundancy
because of the recursive nature of pandoc types, but testing each type
individually will help us pinpoint problems.

Note that this uses a simple quasiquoting library (strings-qq) to avoid
escaping every quote in a JSON string.

### DIFF
--- a/pandoc-types.cabal
+++ b/pandoc-types.cabal
@@ -65,5 +65,12 @@ test-suite test-pandoc-types
   build-depends:       base,
                        pandoc-types,
                        aeson >= 0.6.2 && < 1.1,
-                       QuickCheck >= 2
+                       containers >= 0.3,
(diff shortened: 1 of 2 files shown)


jkr:
sorry -- fixing a couple of -Wall errors

jkr:
okay -- that should fix it.

jgm:
Slightly worried about the quasiquoting -- I've had trouble
with TH before (on Windows), but maybe this is a thing of
the past now.

Anyway, I'll merge this and let's see how it goes.
Thanks!

PR #25, opened 2016-10-10
Add tests for LineBlock elements

tarleb:
LineBlocks are generated with an open-ended number of lines and up to three
inline elements per line.

### DIFF
--- a/Text/Pandoc/Arbitrary.hs
+++ b/Text/Pandoc/Arbitrary.hs
@@ -3,8 +3,8 @@
 -- provides Arbitrary instance for Pandoc types
 module Text.Pandoc.Arbitrary ()
 where
-import Test.QuickCheck 
-import Control.Monad (liftM, liftM2)
+import Test.QuickCheck
+import Control.Monad (forM, liftM, liftM2)
 import Text.Pandoc.Definition
 import Text.Pandoc.Builder
 
@@ -75,6 +75,11 @@ arbBlock :: Int -> Gen Block
 arbBlock n = frequency $ [ (10, liftM Plain $ arbInlines (n-1))
                          , (15, liftM Para $ arbInlines (n-1))
                          , (5,  liftM2 CodeBlock arbAttr realString)
+                         , (3,  liftM LineBlock $
+                                  liftM2 (:)
+                                         (arbInlines $ (n - 1) `mod` 3)
+                                         (forM [1..((n - 1) `div` 3)]
+                                               (const $ arbInlines 3)))
                          , (2,  elements [ RawBlock (Format "html")
                                             "<div>\n*&amp;*\n</div>"
                                          , RawBlock (Format "latex")
@@ -188,4 +193,3 @@ instance Arbitrary ListNumberDelim where
                    2 -> return OneParen
                    3 -> return TwoParens
                    _ -> error "FATAL ERROR: Arbitrary instance, logic bug"
-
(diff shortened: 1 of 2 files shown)


tarleb:
The failed tests are due to non-deterministic ordering of MetaValues and unrelated to the changes at hand.  Not sure how to fix this.

jgm:
@jkr - note the failed test at
https://ci.appveyor.com/project/jgm/pandoc-types

Any thoughts on how to fix this?

+++ Albert Krewinkel [Oct 10 16 14:08 ]:

jkr:
Yep -- I was actually just about to make a pull request that included this (as well as stuff that Albert already did). Easiest answer: knock the MetaMap down to one element. After all, we're not testing whether Aeson's Map instance works. Much harder answer: change the test functions so they encode to a Value instead of a Pandoc, sort within that, and *then* convert to a Meta. There is probably a middle road somewhere in Aeson, but I'd vote for going for the easier one.

I also wanted to add span and div to arbitrary (we missed a bug because they weren't in there). So I could just piggyback this change on that.

jkr:
Oddly this only seems to come up with the MetaMap and not with the "t"/"c" arrays -- I'd have to poke around aeson's internals to see why, but it's so consistent that I tend to think we're safe there.

jgm:
Easy approach is okay for now.

Another idea is that instead of checking raw strings (after
all, strings could differ in many ways for equivalent JSON,
and aeson's behavior may change), we could do the comparison
this way:

Parse both actual and expected value (JSON encoded strings)
into an Aeson Value (using aeson itself -- after all we're
not testing it and can assume that it works properly).
Then use `==` on the Aeson Values.

Even better, in my opinion:  don't encode the expected
results as raw strings (that eliminates the need for
the quasiquoting library). Instead, specify the expected
results using aeson's syntax for building up Values.


+++ Jesse Rosenthal [Oct 11 16 06:05 ]:

jkr:
John MacFarlane <notifications@github.com> writes:
> Another idea is that instead of checking raw strings (after
> all, strings could differ in many ways for equivalent JSON,
> and aeson's behavior may change)...

The problem here is that we're no longer testing the actual JSON
output/input. So if aeson changes anything in the
`Value <--> ByteString` step, we'd miss it. With the current approach,
at least, we're guaranteeing that the JSON that we tell lib writers in
other languages to write/consume will work as expected.

I agree it would make a lot of things easier -- I'm just afraid that we
wouldn't be testing the thing that we most need to test.

jkr:
> I agree it would make a lot of things easier -- I'm just afraid that we
> wouldn't be testing the thing that we most need to test.

On the other hand, if "Value <-> ByteString" is guaranteed (in whatever
sense) to be one-to-one, and there isn't much left up to interpretation,
that might not be too much of an issue.

jgm:
+++ Jesse Rosenthal [Oct 11 16 06:55 ]:
>   > I agree it would make a lot of things easier -- I'm just afraid that
>   we
>   > wouldn't be testing the thing that we most need to test.
>   On the other hand, if "Value <-> ByteString" is guaranteed (in whatever
>   sense) to be one-to-one, and there isn't much left up to
>   interpretation,
>   that might not be too much of an issue.

My thought is that it's aeson's responsibility to correctly
map a Value onto a ByteString, our responsibility is just
to produce the right JSON object.  So we don't need to test
aeson.

I don't feel strongly about this, though -- you're certainly
right that the string is closer to what library writers
will see.

PR #28, opened 2016-11-14
Adds PageBreak definitions

hubertp-lshift:
Blocks https://github.com/jgm/pandoc/pull/3230


hubertp-lshift:
@jgm It looks like those are not legitimate build errors but rather the infrastructure ones.

PR #30, opened 2017-09-11
Default implementation for walk

tarleb:
The implementations for `walk` and `walkM` are very similar, so a default method is provided which implements the former in terms of the latter. This change should not affect performance, as the `Identity` functor, which is used in the default definition, is a newtype that should be eliminated at compile time.

GHC 7.8 (i.e., base <= 4.7) does not provide the `Data.Functor.Identity` type, so we use the one included in the transformers library. The transformers library was already an indirect dependency due to Aeson, so while requiring it is ugly, it does not cause additional code to be included.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+/dist
+/.stack-work
--- a/Text/Pandoc/Walk.hs
+++ b/Text/Pandoc/Walk.hs
@@ -90,6 +90,7 @@ linked to in a document:
 module Text.Pandoc.Walk (Walkable(..))
 where
 import Control.Applicative ((<$>))
+import Data.Functor.Identity (Identity (runIdentity))
 import Text.Pandoc.Definition
 import qualified Data.Traversable as T
 import Data.Traversable (Traversable)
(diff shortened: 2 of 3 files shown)


jgm:
This is a nice simplification.  Are you sure it doesn't have performance ramifications? It would be nice to have some kind of benchmark.

tarleb:
I did a small benchmark using criterion and found a ~10% performance degradation. I don't understand where this comes from. Lesson learned: measuring trumps reasoning.

### Benchmarking code

```haskell
import Criterion.Main (bench, defaultMain, nf)
import Data.Aeson (eitherDecode)
import Text.Pandoc.Definition (Pandoc, Inline (Str))
import Text.Pandoc.Walk (walk)

import qualified Data.ByteString.Lazy as BSL

main :: IO ()
main = do
  pd <- parseManual
  defaultMain [ bench "walk" $ nf (walk prependZeroWidthSpace) pd]

prependZeroWidthSpace :: Inline -> Inline
prependZeroWidthSpace (Str s) = Str ('\8203' : s)
prependZeroWidthSpace x = x

parseManual :: IO Pandoc
parseManual = do
  content <- eitherDecode <$> BSL.readFile "manual.json"
  case content of
    Right pd -> return pd
    _ -> error "couldn't read manual"
```

### Results

#### master
```
benchmarking walk
time                 2.647 ms   (2.620 ms .. 2.678 ms)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 2.658 ms   (2.643 ms .. 2.676 ms)
std dev              52.16 μs   (36.70 μs .. 78.66 μs)
```

#### This PR
```
benchmarking walk
time                 2.931 ms   (2.907 ms .. 2.962 ms)
                     0.999 R²   (0.999 R² .. 1.000 R²)
mean                 2.958 ms   (2.939 ms .. 2.988 ms)
std dev              74.13 μs   (52.64 μs .. 113.0 μs)
variance introduced by outliers: 11% (moderately inflated)
```

jgm:
10% isn't too bad, so we might think about whether the
simplification is worth it.

+++ Albert Krewinkel [Sep 12 17 22:26 ]:

jgm:
I wonder if just adding a SPECIALIZE pragma (for the Identity instance) would close the performance gap?

tarleb:
Adding `{-# SPECIALIZE Identity :: Inline -> Identity Inline #-}` gave me a syntax error, so I manually added `walk f = runIdentity . walkM (return . f) {-# INLINE walk #-}` to all instance definitions, but that didn't help. 🤔

tarleb:
As was pointed out to me by @leftaroundabout, enabling GHC optimizations *mostly* fixes this: my benchmarks now show an increase in runtime of only ~4%.

jgm:
OK, this seems worthwhile, then.

PR #31, opened 2017-09-14
Force optimizations when compiling Walk module

tarleb:
The functions defined in the Walk module are relatively simple while
being called recursively on possibly long documents. It hence seems
sensible to force optimizations for this module, as the benefits likely
outweigh the additional compile time.

This also reduces the performance cost of having a default
implementation for `walk` based on `walkM`.


jgm:
Just wondering why we'd want to make it impossible to disable the optimizations.  For production, one would normally compile with optimizations anyway, right?

jgm:
I see that -O1 is the default for cabal/stack, and thus what we're using for pandoc and pandoc-types, so maybe it makes sense to ensure that this one module is compiled with higher optimization.

PR #32, opened 2017-09-19
Walkable element lists

tarleb:
Add Walkable instances for lists of Inlines and lists of Blocks.  This is a prerequisite for jgm/pandoc#3926.

### DIFF
--- a/Text/Pandoc/Walk.hs
+++ b/Text/Pandoc/Walk.hs
@@ -89,7 +89,7 @@ linked to in a document:
 
 module Text.Pandoc.Walk (Walkable(..))
 where
-import Control.Applicative ((<$>))
+import Control.Applicative (Applicative, (<$>))
 import Data.Functor.Identity (Identity (runIdentity))
 import Text.Pandoc.Definition
 import qualified Data.Traversable as T
@@ -108,7 +108,7 @@ class Walkable a b where
   walk  :: (a -> a) -> b -> b
   walk f = runIdentity . walkM (return . f)
   -- | A monadic version of 'walk'.
-  walkM :: (Monad m, Functor m) => (a -> m a) -> b -> m b
+  walkM :: (Monad m, Applicative m, Functor m) => (a -> m a) -> b -> m b
   -- | @query f x@ walks the structure @x@ (bottom up) and applies @f@
   -- to every @a@, appending the results.
   query :: Monoid c => (a -> c) -> b -> c
--- a/Text/Pandoc/Walk.hs
+++ b/Text/Pandoc/Walk.hs
@@ -89,7 +89,8 @@ linked to in a document:
 
 module Text.Pandoc.Walk (Walkable(..))
 where
-import Control.Applicative (Applicative, (<$>))
+import Control.Applicative (Applicative (pure), (<$>), (<*>))
+import Control.Monad ((>=>))
 import Data.Functor.Identity (Identity (runIdentity))
 import Text.Pandoc.Definition
 import qualified Data.Traversable as T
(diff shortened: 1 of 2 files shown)


tarleb:
No tests have been added yet, as the default way to test instances (`p_walk`) seems to return incorrect results for lists. Running 

```haskell
main :: IO ()
main = toJSONFilter (walk inlinesTrans :: Pandoc -> Pandoc)

inlinesTrans :: [Inline] -> [Inline]
inlinesTrans xs | all whitespaceInline xs = []
  where whitespaceInline Space = True
        whitespaceInline LineBreak = True
        whitespaceInline SoftBreak = True
        whitespaceInline (Str "") = True
        whitespaceInline (Span _ []) = True
        whitespaceInline _ = False
inlinesTrans x = x
```

on

    Pandoc (Meta {unMeta = []}) [Para [Str "a", Str ""]]

gives

    Pandoc (Meta {unMeta = []}) [Para "a"]

where `walk` instead of `bottomUp` leaves the document unchanged.

I'm not sure what's the best way to test these new instances.

jgm:
+++ Albert Krewinkel [Sep 19 17 15:24 ]:
>   No tests have been added yet, as the default way to test instances
>   (p_walk) seems to return incorrect results for lists. Running

Could you say more about how?

Note that bottomUp (as used, definitionally expanded, in
p_walk) gets applied to every sublist of a list:

```
Prelude Text.Pandoc.Generic Text.Pandoc.Definition> let f (xs :: [Inline]) = [Emph xs]
Prelude Text.Pandoc.Generic Text.Pandoc.Definition> bottomUp f $ [Para [Str "hi", Space, Str "there", Space, Str "friend"]]
[Para [Emph [Str "hi",Emph [Space,Emph [Str "there",Emph [Space,Emph [Str "friend",Emph []]]]]]]]
```

I don't know if your walk instance is designed to work like
that, or differently?

jgm:
See https://stackoverflow.com/questions/2428161/lists-searches-in-syb-or-uniplate-haskell
for the basic issue.  The solution there uses uniplate.  Perhaps something similar could be done with syb?

jgm:
Some thoughts:

(1) One way to remove duplicate code would be to factor out `walkBlockM` and `walkInlineM` to the top level, then use these to define `walkM` instances for both `a -> a` and `[a] -> [a]` forms:

```haskell
instance Walkable Inline Inline where
  walkM f = walkInlineM >>= f

instance OVERLAPS
         Walkable [Inline] [Inline] where
  walkM f = traverse walkInlineM >=> f
```
I haven't checked this completely, but at first glance it looks like this would work.  That would fix the code duplication problem.

(2) There is a travis failure for ghc 7.8, which will require some extra imports (possibly protected by CPP so we don't get warnings for other versions).

(3) I'm wondering whether we really need this feature, and the complexities it brings.  I see from the associated lua filter changes that we'd be, essentially, lifting a function Inline -> [Inline] to a function [Inline] -> [Inline] by using map and mconcat.  That means that *every* filter will incur the overhead of concatenating a list of lists (generally, one element lists).  I don't know if you've measured the performance impact of this, but I expect it would be significant.

We could keep things simple and have lua filters define Inline -> Inline or Block -> Block.  To delete an element, convert it to a Null block or `Str ""`.  To insert a list of elements, insert a span.  (In most formats, this will be no different from splicing a list, if the span has no attributes; in HTML, you'll get a span tag, but that's harmless.)

On the other hand, there are some uses for filters that operate on [Inline] -> [Inline]; in pandoc-citeproc, I have to resort to `bottomUp` for the function to move punctuation marks (mvPunct in Text.CSL.Pandoc); this Walk instance might be faster.  Not sure that justifies all the additional complexity, though.

tarleb:
(1) This was my approach as well, but didn't work the way I expected. The type signature for `walkInlineM` would have to be along the lines of

    walkInlineM :: (…) => (t Inline -> m (t Inline)) -> (t Inline) -> m (t Inline)

where `t` would be specialized to `[]` or `Identity`.  However, this requires to change the `Walkable [Inline] [Inline]` to `Walkable (t Inline) [Inline]`, which I either found to require InconsitentInstances or just couldn't do (I don't remember).

(3) I checked the performance impact beforehand, and (contrary to my own assumptions) using the `Walkable [Inline] Pandoc` instance does not significantly affect performance.

I don't have a clear opinion on whether it's worth the additional complexity. I currently have a slight preference for adding those instances but remain sceptical.

tarleb:
I used the assumption

```haskell
everywhere (mkT f) d == walk (foldr g []) d
  where g x ys = f (x:ys)
```

to test the implementations for `walkM` and did find some issues here and there.

There seems to be no such law which would make it possible to fully test `query`, so I'm currently only testing that the function given to query gets to see the correct amount of elements.

jgm:
I wonder if it would make more sense to have the [Inline] -> [Inline] instance work the way bottomUp does (operating on each sublist of a list)?  Then we would match the behavior of Text.Pandoc.Generics (least surprise) and could test with that.  We could also sub out some existing uses of bottomUp without worrying about different semantics.

Would it be difficult to change the code to do this?  Would it make a difference for the application to lua filters? (Presumably not since you're always passing in a one-element list.) Would it affect performance?

Example of how this works in practice:
```
Prelude Text.Pandoc.Definition Text.Pandoc.Generic> let f [Str "."] = []; f xs = xs
Prelude Text.Pandoc.Definition Text.Pandoc.Generic> bottomUp f [Str "hi", Str ".", Str "there", Str "."]
[Str "hi",Str ".",Str "there"]
```

jgm:
I merged this (though I may still want to revert of modify it).
What do you think of the alternative above?

I also added a commit that removes ALL the code duplication.
Let me know if there's something bad about this that I've missed!

jgm:
I pushed another commit that makes walkM and query match behavior of bottomUp and queryWith for [Inline] -> [Inline] and [Block] -> [Block].  Let me know if this is a bad idea!

tarleb:
Oh wow. This is great (and I'm feeling very stupid now)!

tarleb:
Performance wise, having walkM match bottomUp is very costly. It turns an O(n) operation into an O(n²) operation, which is very noticable with lua filters. Furthermore, counting elements with lua filters is no longer possible, as the number of times a function is called with a given element is dependent on its context.

tarleb:
There is also the mathematical argument, that the `Walkable [a] [a]` implementations for walkM are the only ones which require the functor `m` to be a monad. All other instances could be expressed with `m` being Applicative. Weakening the constraint for walkM would bring its type closer to that of a lens [`Traversal'`](https://www.stackage.org/haddock/lts-9.5/microlens-0.4.8.1/Lens-Micro-Type.html#t:Traversal-39-), which might become useful at some point.

tarleb:
nm my previous comment. I should step away from the computer for a while.

jgm:
> Performance wise, having walkM match bottomUp is very costly. It turns an O(n) operation into an O(n²) operation, which is very noticable with lua filters.

I was thinking this wouldn't matter for the particular application, since n = 1 (you're passing in singleton lists).

> . Furthermore, counting elements with lua filters is no longer possible, as the number of times a function is called with a given element is dependent on the element's context.

Yes, I see the concern there.

jgm:
OK, I reverted that last change.

PR #34, opened 2017-11-05
Add note type and functions to create endnotes

labdsf:


labdsf:
@jgm Can you please review this PR and stamp a version so I can update jgm/pandoc#4042?

aparcar:
Any progress on that?

labdsf:
@aparcar You can compile my ["endnotes" branch](https://github.com/labdsf/pandoc/tree/endnotes) and test it. I keep it updated by rebase it on top of jgm/pandoc master.

I am in the process of adding Lua Note constructor compatibility and tests. After that I think it will be ready for merge.

labdsf:
@aparcar 
> What is required to add EndNote support to markdown, as that the file format I'm currently working on?

There is a PR #1425 with a discussion on markdown endnote syntax. The syntax for endnotes in Markdown is not decided, so I have not added endnote support to markdown reader.

If you want to keep using Markdown, you can easily extend it with a Lua filter. Take a look at `test/lua/note.lua`, which unconditionally replaces all footnotes with endnotes. You can add some condition to distinguish between footnotes that you want to turn into endnotes and those you want to keep being footnotes.

aparcar:
It's okay to have only EndNotes so I added `--lua-filter=../note.lua` which results in a lot of "ok" and "FootNote" printing, however, the final documents stays the same. Anything else to look at?

aparcar:
Great, with `odt` it does work! 

Would it be possible to a] support PDF as output and b] to use numbers instead of roman numbering? Should I open additional issues for both?

labdsf:
@aparcar I have already removed that "ok" debug printing, you need to update to remove it.

For PDF output it is necessary to add endnote support to LaTeX writer. LaTeX does not have native support for endnotes, so it will be necessary to decide which package to use for that and things like that. 
If you find ODT typography satisfactory, you may convert to ODT and then convert to PDF with headless LibreOffice. Another option is to convert to Muse and then use Text::Amuse::Compile Perl package to convert it to to LaTeX and PDF.

Roman numbering seems to be the default ODT style, you can probably change it somehow with templates. Text::Amuse::Compile uses numbers by default, so you will like it.

aparcar:
@labdsf thank you very much for your kind help! I'd rather stick with the LaTeX typography, who complex is the process of integrating an external library package?

labdsf:
@aparcar I think this discussion is a bit off-topic for this PR. Just one more remark: if you just want to move all your footnotes to the end of the document, you may just change your LaTeX template by adding `\let\footnote=\endnote` as suggested here: https://tex.stackexchange.com/questions/56145/is-there-a-way-to-move-all-footnotes-to-the-end-of-the-document

labdsf:
@jgm Is there any plan to get this merged? All the feedback for https://github.com/jgm/pandoc/pull/4042 is addressed, I need a new version with this PR merged and new version of pandoc-types released to finalize it.

PR #35, opened 2017-11-22
Consequently use the `Semigroup` class for `<>`.

leftaroundabout:
This will actually become necessary when Monoid is a subclass of Semigroup.

https://ghc.haskell.org/trac/ghc/ticket/14191

### DIFF
--- a/Text/Pandoc/Builder.hs
+++ b/Text/Pandoc/Builder.hs
@@ -166,7 +166,7 @@ module Text.Pandoc.Builder ( module Text.Pandoc.Definition
 where
 import Text.Pandoc.Definition
 import Data.String
-import Data.Monoid
+import Data.Semigroup
 import qualified Data.Map as M
 import Data.Sequence (Seq, (|>), viewr, viewl, ViewR(..), ViewL(..))
 import qualified Data.Sequence as Seq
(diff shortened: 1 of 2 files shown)


leftaroundabout:
Note: this breaks the pandoc build, baring https://github.com/jgm/pandoc/pull/4089

jgm:
I'm not sure this is the right time for this change.  Presumably when this happens, semigroup will become part of base. At that point, we'll need further adjustments, since the semigroup package will not be needed for base versions that include it.  Is there a compelling reason to make this change now rather than waiting for ghc 8.4?

leftaroundabout:
The `Semigroup` module [is _already_ in `base`](http://hackage.haskell.org/package/base-4.9.0.0/docs/Data-Semigroup.html). Thus, the [`semigroups` package](http://hackage.haskell.org/package/semigroups) isn't actually needed as of GHC-8.0 / 8.2, nor will it be needed in the future; but by depending on `semigroups` we can ensure that the build works just the same for older GHC version.

Even with the GHC<8.4 class hierarchy, it is a problem for some applications if there's no `Semigroup` instance, because they may use that constraint, so right now we need some ugly `WrappedMonoid` trickery when using Pandoc types with such functions.

PR #36, opened 2017-11-23
pad table headers up to max row length to avoid syntax errors, closes…

danse:
… #4059

### DIFF
--- a/Text/Pandoc/Builder.hs
+++ b/Text/Pandoc/Builder.hs
@@ -471,21 +471,19 @@ headerWith attr level = singleton . Header level attr . toList
 horizontalRule :: Blocks
 horizontalRule = singleton HorizontalRule
 
+-- | Table builder. Rows and headers will be padded or truncated to the size of
+-- @cellspecs@
 table :: Inlines               -- ^ Caption
       -> [(Alignment, Double)] -- ^ Column alignments and fractional widths
       -> [Blocks]              -- ^ Headers
       -> [[Blocks]]            -- ^ Rows
       -> Blocks
 table caption cellspecs headers rows = singleton $
-  Table (toList caption) aligns widths
-      (map toList headers') (map (map toList) rows)
+  Table (toList caption) aligns widths (sanitise headers) (map sanitise rows)
    where (aligns, widths) = unzip cellspecs
-         numcols  = case (headers:rows) of
-                         [] -> 0
-                         xs -> maximum (map length xs)
-         headers' = if null headers
-                       then replicate numcols mempty
-                       else headers
+         sanitise = map toList . pad mempty numcols
+         numcols = length cellspecs
+         pad element upTo list = take upTo (list ++ repeat element)
 
 -- | A simple table without a caption.
 simpleTable :: [Blocks]   -- ^ Headers
--- a/test/test-pandoc-types.hs
+++ b/test/test-pandoc-types.hs
@@ -3,6 +3,7 @@
 import Text.Pandoc.Arbitrary ()
 import Text.Pandoc.Definition
 import Text.Pandoc.Walk
+import Text.Pandoc.Builder (simpleTable, singleton)
 import Data.Generics
 import Data.List (tails)
 import Test.HUnit (Assertion, assertEqual, assertFailure)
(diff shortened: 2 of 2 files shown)


Text/Pandoc/Builder.hs
```diff
@@ -483,9 +493,7 @@ table caption cellspecs headers rows = singleton $
          numcols  = case (headers:rows) of
                          [] -> 0
                          xs -> maximum (map length xs)
-         headers' = if null headers
-                       then replicate numcols mempty
-                       else headers
+         headers' = pad mempty numcols headers
```
mb21: Since the `pad` function is only used here, I would move it here: on the same level as `headers'` and `numcols`.

pandoc-types.cabal
```diff
@@ -1,5 +1,5 @@
 Name:                pandoc-types
-Version:             1.17.3
+Version:             1.17.4
```
mb21: jgm will increment the version number in a separate commit when he releases a new version... no need to do it in a pull request. (same for the changelogs AFAIK)

Text/Pandoc/Builder.hs
```diff
@@ -305,6 +305,16 @@ setDate = setMeta "date"
 
 -- Inline list builders
 
+-- | Extend a list with a default element up to a maximum length
```
mb21: we aren't using `doctest` in this project... so I'd just add a few more tests to `test-pandoc-types.hs`

danse:
okay i added a commit with the changes you mentioned, ready to be squashed

jgm:
If I understand correctly, this pads headers to max row length, but doesn't pad rows. Should rows be padded too if they have fewer cells than other rows or the header?

danse:
yeah padding rows is a low-hanging fruit here, i'll add a test and the feature

danse:
i added row padding. some tests are failing but the error seems independent from the changes i made

jgm:
If I'm not mistaken, nothing prevents having a different number of columns in the body than in the `cellspec` (column widths and alignments).  Is this something we should also sanitize?

One option would be to take the cellspec as determining the number of columns, and pad or truncate all rows to that number.

danse:
i'd say to pad also the specification. about the idea to use the specification as an indicator of the max size, i'd rather keep the current max length. i think that it's more sensible because:

- the specification is metadata and the user does not see it on some source formats
- picking the max size, no data is lost

jgm:
Padding the specification is a bit tricky, though.  What if
they specify two nonzero widths, but have a row with three
columns?  Then we'll get a column with 0, and that will
screw up renderers, which expect either all 0s or all
nonzero values.

+++ Francesco Occhipinti [Dec 03 17 17:41 ]:

danse:
if this is getting tricky, i'm not sure that we should keep discussing in this context. the original motivation of this pull request is to address issue #4059 which caused syntax errors to an user. from the point of view of that user, this pull request is already solving a problem. we also thought to extend this to rows, but now it seems like our desire to improve more is preventing us to improve the experience of that user

jgm:
I don't think my proposal was tricky.  The proposal is
simple:  the 'table' builder takes the cellspecs to
determine the number of columns, and pads or truncates
the rows to ensure that number of columns.  Presumably
it's a programming error anyway if someone includes
a row with more cells than the cellspecs.

With this change, we'd have a guarantee that any table
created with the table builder is consistent.

danse:
oh, i see, you were making a point about your proposal while i thought that we were stuck. let me reconsider more carefully

danse:
i updated the pull request, now it pads or truncate depending on `colspecs`. i hope i got the Haddock right :)

danse:
any problems with this? [once merged this will enable pandoc's translation to become more robust to malformed input](https://github.com/jgm/pandoc/issues/4059)

danse:
i tried to run pandoc tests using this version of `pandoc-types` and 49 tests fail. many if not all failures could be due to the change of behaviour, but there could also be some error. i will propose this again when i will have a pull request ready for pandoc tests as well

danse:
alright it was not that bad, three tests were actually failing and [here](https://github.com/italia/pandoc/commit/c2c6e0b6a248fe69d5920721086e45b5e9eba3ac) is the fix.

- in Writers.Muse, empty specs caused truncation of the whole table content
- in 2649, what seems to be a parsing or building error is propagated to all rows
- in 3337, a table is sliced according to the specs the parser is probably passing to the builder

we have a problem similar to 3337 with the DOCX parser, i think that i will change the logic there so that the maximum width is used for the specs

danse:
i saw that there were some updates on `master` so i rebased this. this is ready as far as i can tell

jgm:
Thanks!

danse:
Thanks to you! Let me know whether i can help updating the tests in pandoc. I remind you that some info about how to update them and a commit with the changes are [in the comment above](https://github.com/jgm/pandoc-types/pull/36#issuecomment-364909184)

jgm:
OK, I started looking at pandoc tests.

In command test 3337, we have a table with no header and two body rows, one with two cells, one with one.  The test output expects the table to be padded to two cells, but with this commit it's padded to one cell.

I'm actually starting to change my mind about the decision above; maybe padding to the maximum width makes the most sense?  After all, that is what HTML will do.

jgm:
OK, looking back at my reasons above, they seem sound.
Maybe this can be handled with a change to the HTML reader.

jgm:
I'm going to make the needed changes in pandoc tests and pandoc.

danse:
sorry for being late on this. In the case of 3337, as you said, the HTML reader is responsible for calling the builder with specs that will not lead to any data loss. Thanks for moving this forward, i'll verify that it closes #4059

PR #43, opened 2018-04-10
Add extra imports to fix build on older GHCs

gwils:
As a hackage trustee, I've made revisions to versions 1.17.4.2 and 1.17.4.1 to increase their base lower bounds to `>= 4.8` and `>=4.9` respectively.

This patch will make pandoc-types compile as far back as GHC 7.4

### DIFF
--- a/Text/Pandoc/Arbitrary.hs
+++ b/Text/Pandoc/Arbitrary.hs
@@ -4,6 +4,7 @@
 module Text.Pandoc.Arbitrary ()
 where
 import Test.QuickCheck
+import Control.Applicative (Applicative ((<*>), pure), (<$>))
 import Control.Monad (forM)
 import Text.Pandoc.Definition
 import Text.Pandoc.Builder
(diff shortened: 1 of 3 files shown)


jgm:
I suppose this is innocuous, but I object to your modifying pandoc on Hackage without consulting with me first.  I'm an active maintainer.  And we made a conscious decision to drop support for older GHC versions in pandoc to simplify the maintenance overhead.  (Pandoc used to support these versions of ghc; it's been around since GHC versions started with 6!)  I don't really like the idea of a third party countermanding this intentional decision without consulting first!  In any case, pandoc itself will not be supporting these older versions, but I suppose there may be some other consumers of pandoc-types that could benefit from this.

jgm:
Oh, excuse me, I misunderstood what the Hackage changes were -- I see now they were raising, not lower the base version bounds!  Apologies!

jgm:
Thanks for the PR.  Should I set the bound `base >= 4.5` now? (You said GHC 7.4.)

gwils:
`deepseq-generics` does not have any releases that support `base <4.5`, so it's It's not exactly necessary to set `base >= 4.5` in pandoc-types, but I think it's a good idea to do so, yes.

jgm:
I'll do that. Thanks again!

gwils:
You're welcome :)

PR #47, opened 2018-08-22
add a `mappendPreserve` function in the Builder

danse:
this function can be useful when parsing code blocks, see jgm/pandoc#4841


danse:
see usage in https://github.com/jgm/pandoc/pull/4852

danse:
as discussed in jgm/pandoc#4852, we prefer to change the specific parses rather than shared functions in `pandoc-types`

PR #50, opened 2018-09-10
Flatten inlines and blocks during shrinking

link2xt:


link2xt:
Have to use those fromList/toList functions to make sure shrinking is done using my own functions instead of generic list shrinking that does not flatten the lists.

Is there any plan to replace lists of `Inline` and `Block` with `Inlines` and `Blocks` in the future? Sure it breaks compatibility but allows to use `Seq` directly instead of lists, and can easily be upgraded to any other data structure later.

link2xt:
> I don't really understand what these changes do. Can you give an example?

For example, for a document like
```
[Para [Str "foo",Space,Emph [Str "bar",Space,Str "baz"]]]
```
it will try to reduce it to
```
[Para [Str "foo",Space,Str "bar",Space,Str "baz"]]
```
by flattening the Emph.

Normal shrinking of lists tries to remove elements from the list and shrink each of its element. For example, `Str s` can only be shrinked by removing characters from `s`, and `Emph inls` can only be reduced to `Emph []` but no further. I also try to flatten the list, such as by moving the elements out of `Emph`. Bullet lists are flattened to paragraphs with each bullet content, tables are flattened to paragraphs with their cell content etc.

Without flattening, if you have a bug triggering inline list that is inside a table inside a bullet list inside a definition list, you will never remove those tables and lists, even though they are not necessary to reproduce the bug.

> For purposes like shrinking, you can always convert `[Inline]` to `Inlines` and back, can't you?

That is exactly what I am doing with `shrinkInlineList`, `shrinkInlinesList`, `shrinkBlockList` and `shrinkBlocksList`.

jgm:
So, am I understanding correctly that QuickCheck
applies these various "shrinking" transformations
until the test no longer fails, and gives you the
shortest one that fails?

link2xt:
> So, am I understanding correctly that QuickCheck applies these various "shrinking" transformations until the test no longer fails, and gives you the shortest one that fails?

`shrink` generates a list of reduced test cases given a failing test case. QuickCheck tests all of them, and if it finds a failing one, it replaces the current one with this reduced version. Then it repeats the shrinking, until none of the reduced versions are failing.

It is not necessarily the shortest one, because QuickCheck does not explore the whole tree of transformations, but not immediately reducible (as opposed to SmallCheck, for example, see https://ro-che.info/articles/2018-05-25-quickcheck-vs-smallcheck).

By default `shrink` returns `[]` for custom types, so shrinking only works for them if you define a custom version of it. That is what I did.

PR #51, opened 2018-09-16
Definition.hs: add and export new lookupMeta* functions

mb21:
`lookupMetaInlines`, `lookupMetaBlocks` and `lookupMetaBool`

This is an API change.

### DIFF
--- a/Text/Pandoc/Definition.hs
+++ b/Text/Pandoc/Definition.hs
@@ -52,6 +52,9 @@ module Text.Pandoc.Definition ( Pandoc(..)
                               , nullMeta
                               , isNullMeta
                               , lookupMeta
+                              , lookupMetaBool
+                              , lookupMetaBlocks
+                              , lookupMetaInlines
                               , docTitle
                               , docAuthors
                               , docDate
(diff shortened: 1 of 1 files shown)


jgm:
Instead of having these different functions, we could introduce a version of `lookupMeta` with a more general type:
```
lookupMeta' :: FromMetaValue a => Meta -> Maybe a
```
and have a `FromMetaValue` class.  That might be more elegant, no?

mb21:
Hm.. since none of the three added functions return a `Maybe`, it would have to be:

    lookupMeta' :: FromMetaValue a => String -> Meta -> a

But then we'd need to enable [`FlexibleInstances`](https://wiki.haskell.org/List_instance). Or we go with:

    lookupMeta' :: FromMetaValue a => String -> Meta -> [a]

But that leaves the `Bool` case uncovered.

mb21:
Or I can simply add the three new functions to `Text.Pandoc.Writers.Shared` instead. Although they are indeed very similar to those existing in pandoc-types: `lookupMeta`, `docTitle`, `docAuthors` and `docDate`.

jgm:
Yeah, let's do it that way, at least for now.

mb21:
Okay, I'll leave this pull open in case you want to move the functions over from `Text.Pandoc.Writers.Shared`.

PR #53, opened 2018-12-08
Added underline support 

dtheras:
Added an inline constructor for native underline support. To close pandoc issue #5044.


jgm:
obsoleted by #68

PR #54, opened 2019-01-07
Change homepage to pandoc.org

chris-martin:
The URL currently listed as the homepage just redirects to `pandoc.org`.


jgm:
Thanks!

PR #56, opened 2019-06-15
Use enumerable set of known formats

tarleb:


tarleb:
@mb21 noted that we might want to enforce a stricter separation between formats and output formats, which is not the case in this PR.

mb21:
@tarleb Yes, we probably should use the rule of thumb that a `Format` corresponds to a writer. That would mean there is `markdown`, but not `markdown_mmd`. There is `HTML`, but no `HTML4`, `EPUB`, `slidy`, etc. That makes sure that we produce syntactically valid output. (And people can always write different filters if they want different behaviours for different outputs.)

The exception to this rule is of course `tex`, which is still a format even though we have a `latex` and a `context` writer.

jgm:
One thought I've sometimes had is that it would be cool if you could specify disjunctions of formats:
```
Format (LaTeX `or` ConTeXt)
```
or maybe even complements:
```
Format (not (LaTeX `or` ConTexT))
```

tarleb:
I added experimental support for the `IfFormat...` constructors suggested in jgm/pandoc#547

tarleb:
Note on why this is stalling (as far as I remember): it's not always clear what a "Format" really is, and I had trouble finding a good abstraction to capture this.

Format seems to be used as

- a specific format, e.g. `icml`, `rst`;
- a descriptor of a class of formats, e.g. `tex`, `html`, `markdown`;
- a specific variant of a format variant, e.g. `html5`, `epub2`, `latex`, `context`.

Some of this can be handled by making `Format` a collection of formats, e.g. `html` being equivalent to `[html4, html5]`. I recall two issues with this:

(1) Sometimes a generic format has be be selected, e.g., 'html5' when `html` is specified as an input format. 
(2) the direction of a subset-relation is unclear. Is `epub ⊂ html` because EPUB is an HTML-based format, or is `html ⊂ epub` as raw HTML can occur in EPUB output.

All this isn't properly captured in the current PR, it has to be rewritten.

jgm:
Maybe the Format type has to be a partial order. That would help with 1 but not 2.

PR #57, opened 2019-08-06
Fix compiler and hlint warnings.

peteryland:


peteryland:
This is not passing CI, I'll look into it.

peteryland:
I've also added a GHC 8.6 build to the Travis matrix.

jgm:
OK, thanks!

PR #58, opened 2019-08-08
Walk: export walk and query helpers

tarleb:
The `walk*M` and `query*` functions are helpful when defining new
`Walkable` instances.

### DIFF
--- a/Text/Pandoc/Walk.hs
+++ b/Text/Pandoc/Walk.hs
@@ -136,6 +136,7 @@ instance (Foldable t, Traversable t, Walkable a b) => Walkable a (t b) where
   walkM f = T.mapM (walkM f)
   query f = F.foldMap (query f)
 
+-- Walk pairs by handling both elements, then combine the results.
 instance OVERLAPS
         (Walkable a b, Walkable a c) => Walkable a (b,c) where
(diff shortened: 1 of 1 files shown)


jgm:
If we're exporting these, it would be good to add Haddock comments documenting them (and perhaps indicating somehow that they're intended for defining new instances rather than for normal use?).

Question:  If the reason we need to export this is to support the instance you define in the lua module, which causes ordered traversal, should we perhaps consider instead just changing the default instances so they do this?  (Then we wouldn't need to export these helpers.)

tarleb:
Ah, the docs. I'll write some.

I have mixed feelings about changing the `Walkable` instances. I don't think the instance is used anywhere but pandoc, so this should be fine. On the other hand, I feel like the new `[Inline]` instance would be against the spirit of Walkable in that it would be geared towards our filter use-case, and would breaks some reasonable expectations. E.g., if I was to replace all inline lists consisting of just a single image and were to use Walkable, I might find that it replaces *all* images, as all list elements are passed to the filter as singletons at some point of the traversal.

I guess I see the new traversal method as kind of a hack. As such, I'd rather hide it away in an unexported module.

There is also the question of what such a change would mean for versioning. I find the PVP inconclusive in that regard.

PR #61, opened 2019-10-30
Use Text instead of String internally

despresc:
In the projects [issue](https://github.com/jgm/pandoc/issues/5581) in the main pandoc repository, this change was mentioned as a possible enhancement. I thought that I should start here and submit this pull request for comment before moving on to the other packages.

There is one place where I was unsure: the new `text` dependency currently lacks a version bound. The functions that are needed existed in that package as early as `text-0.3`, but I was not sure if such a generous bound was appropriate.

The only observable difference in behaviour introduced by these changes relates to `Format`, whose instances now use the `Data.Text.toCaseFold` instead of `map Data.Char.toLower`. Both the former function and `Data.Text.toLower` behave differently than the latter function, and `Data.Text.toCaseFold` is mentioned in `Data.Text` as being intended for case-insensitive comparison, though it does differ a little more in behaviour than `Data.Text.toLower`. This should not affect any of the recognized `Format` names, unless there is an exotic `Format` name that I am unaware of.

### DIFF
--- a/Text/Pandoc/Arbitrary.hs
+++ b/Text/Pandoc/Arbitrary.hs
@@ -15,9 +15,11 @@ realString :: Gen Text
 realString = fmap T.pack $ resize 8 $ listOf $ frequency [ (9, elements [' '..'\127'])
                                                          , (1, elements ['\128'..'\9999']) ]
 
-instance Arbitrary Text where
-  arbitrary = T.pack <$> arbitrary
-  shrink xs = T.pack <$> shrink (T.unpack xs)
+shrinkText :: Text -> [Text]
+shrinkText xs = T.pack <$> shrink (T.unpack xs)
+
+shrinkText2 :: (Text, Text) -> [(Text, Text)]
+shrinkText2 = liftShrink2 shrinkText shrinkText
 
 arbAttr :: Gen Attr
 arbAttr = do
@@ -25,6 +27,12 @@ arbAttr = do
   classes' <- elements [[],["haskell"],["c","numberLines"]]
   keyvals <- elements [[],[("start","22")],[("a","11"),("b_2","a b c")]]
   return (id',classes',keyvals)
+
+shrinkAttr :: Attr -> [Attr]
+shrinkAttr (a, b, c)
+  = [ (a', b', c') | a' <- shrinkText a,
+                     b' <- liftShrink shrinkText b,
+                     c' <- liftShrink shrinkText2 c ]
 
 instance Arbitrary Inlines where
   arbitrary = (fromList :: [Inline] -> Inlines) <$> arbitrary
@@ -91,7 +99,7 @@ shrinkBlocksList = fmap (fmap toList) . shrink . fmap fromList
 
 instance Arbitrary Inline where
   arbitrary = resize 3 $ arbInline 2
-  shrink (Str s) = Str <$> shrink s
+  shrink (Str s) = Str <$> shrinkText s
   shrink (Emph ils) = Emph <$> shrinkInlineList ils
   shrink (Strong ils) = Strong <$> shrinkInlineList ils
   shrink (Strikeout ils) = Strikeout <$> shrinkInlineList ils
@@ -101,22 +109,22 @@ instance Arbitrary Inline where
   shrink (Quoted qtype ils) = Quoted qtype <$> shrinkInlineList ils
   shrink (Cite cits ils) = (Cite cits <$> shrinkInlineList ils)
                         ++ (flip Cite ils <$> shrink cits)
-  shrink (Code attr s) = (Code attr <$> shrink s)
-                      ++ (flip Code s <$> shrink attr)
+  shrink (Code attr s) = (Code attr <$> shrinkText s)
+                      ++ (flip Code s <$> shrinkAttr attr)
   shrink Space = []
   shrink SoftBreak = []
   shrink LineBreak = []
-  shrink (Math mtype s) = Math mtype <$> shrink s
-  shrink (RawInline fmt s) = RawInline fmt <$> shrink s
+  shrink (Math mtype s) = Math mtype <$> shrinkText s
+  shrink (RawInline fmt s) = RawInline fmt <$> shrinkText s
   shrink (Link attr ils target) = [Link attr ils' target | ils' <- shrinkInlineList ils]
-                               ++ [Link attr ils target' | target' <- shrink target]
-                               ++ [Link attr' ils target | attr' <- shrink attr]
+                               ++ [Link attr ils target' | target' <- shrinkText2 target]
+                               ++ [Link attr' ils target | attr' <- shrinkAttr attr]
   shrink (Image attr ils target) = [Image attr ils' target | ils' <- shrinkInlineList ils]
-                                ++ [Image attr ils target' | target' <- shrink target]
-                                ++ [Image attr' ils target | attr' <- shrink attr]
+                                ++ [Image attr ils target' | target' <- shrinkText2 target]
+                                ++ [Image attr' ils target | attr' <- shrinkAttr attr]
   shrink (Note blks) = Note <$> shrinkBlockList blks
   shrink (Span attr s) = (Span attr <$> shrink s)
-                      ++ (flip Span s <$> shrink attr)
+                      ++ (flip Span s <$> shrinkAttr attr)
 
 arbInlines :: Int -> Gen [Inline]
 arbInlines n = listOf1 (arbInline n) `suchThat` (not . startsWithSpace)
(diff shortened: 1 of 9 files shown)


tarleb:
Two ideas which could make transitioning of dependent packages easier:

- Create a `Text.Pandoc.Definitions.Legacy` module, which behaves just like the current `Definitions`. Use pattern synonyms to go from `Legacy.Str` to `New.Str` etc.
- Similarly, add a `Legacy.Builder` module.

This should make transitioning as simple as replacing import statements, allowing to switch to Text module by module, instead of having to do it all at once.

jgm:
> This should make transitioning as simple as replacing import statements, allowing to switch to Text module by module, instead of having to do it all at once.

One approach for doing this could be to define parameterized types: `Pandoc' a`, `Block' a`, `Inline' a`, etc.  Locally we could define `Pandoc = Pandoc String` etc. in unconverted modules and `Pandoc = Pandoc Text` in converted ones.

But I think this would mainly be useful in the transition.  If the transition could be done quickly enough, it might not be necessary.

jgm:
Thinking on this some more, @tarleb's suggestion is probably better.
@tarleb, can you clarify how pattern synonyms could be used?
Alternatively, the legacy module could simply provide a function that converts between old-style and new-style AST.

As for builder, is it clear we need a new interface?

Currently the builder constructors `text`, etc. takes strings as arguments.  Suppose we change this so it takes text.  Then the user needs to use `OverloadedStrings`.  And this also introduces an ambiguity, since we have an `IsString` instance for `Inlines`: `"hi"` could be a Text, a String, or an Inlines (equivalent to `text "hi"`).  So more type signatures might be needed.  Need to think about how to balance the ergonomics here.  I can imagine it being strange and annoying if Builder continues to use String, but there are some costs.

despresc:
I think the suggestion is to define something like

```.haskell
{-# LANGUAGE PatternSynonyms, ViewPatterns #-}

module Text.Pandoc.Legacy.Definition 
  ( D.Inline -- No constructors exported
  , pattern Str
  , <many more patterns>
  ) where

import qualified Text.Pandoc.Definition as D
import qualified Data.Text as T

pattern Str :: String -> D.Inline
pattern Str x = D.Str (T.unpack -> x) where
  Str x = D.Str $ T.pack x
```

Then anyone who used the old `Text.Pandoc.Definition` directly could adapt to this new one by changing all of their imports to `Text.Pandoc.Legacy.Definition` and the existing code will compile and work as before. Notably, users of the module would not have to enable those language extensions.

despresc:
You are right that there is ambiguity in the use of `Builder` and `OverloadedStrings`. The expression `toMetaValue "string"` needs annotation in the presence of `OverloadedStrings`, since it requires `IsString` and `ToMetaValue` to cooperate in instance resolution. Notably, that issue seems to be present in the current release of `Builder` as well. I could not see other sources of ambiguity offhand, but that sort of thing does have a way of happening with `OverloadedStrings`.

I would prefer to change the `Builder` interface to use `Text` by default and find a solution to the ambiguity problem (or tolerate the signatures), since the future use of `Text` will encourage users to use `OverloadedStrings` more often and in situations where this sort of thing will crop up. 

Two immediate issues are the use of the `Attr` type in `Builder` functions, as well as the use of `Citation` in the `cite` builder, since these require `Text` to be constructed. If `Builder` still uses `String`, then these could be solved by providing a `citation :: Text -> ...` builder and having the `*With` builders take `type Attr' = (String, [String], [(String, String)])` arguments instead.

Incidentally, the `ToJSONFilter` class has a `[String] -> a` instance. Should this become `[Text] -> a`?

tarleb:
Thanks @despresc, that's exactly what I had in mind. If finely grained imports are a goal, then using one submodule per data type could make sense, as that seems difficult to control otherwise. So `import Text.Pandoc.Definitions (Inline (..))` would be equivalent to `import Text.Pandoc.Definitions.Legacy.Inline`. But maybe that is not actually needed.

despresc:
It would not be much more difficult to have `Legacy.Definition.*` (or `Definition.Legacy.*`) modules for each of `Meta`, `MetaValue`, `Block`, `Inline`, and `Citation` (that should be all of them) and have `Legacy.Definition` re-export all of them.

If these modules are not intended for a release version, then splitting them up should not be necessary. I can comment out any existing `Definition` import lists (I have seen a few in `pandoc`), add new ones adapted to `Legacy.Definition`, then change each module and restore its old import list. If they are intended for release (and later deprecation and removal), then separate modules might be welcome.

despresc:
I did not know that record pattern synonyms were introduced after GHC 7.10.3. Those are required for complete compatibility, in case record syntax is used with `Citation` or `Meta`, and are what seem to be causing the CI failure. If these modules are intended for wider compatibility, then either the record pattern synonyms will have to be removed or GHC 7.10.3 support will need to be dropped. Otherwise they can stay, as long as continuous integration failures can be temporarily tolerated.

I mentioned a `[Text] -> a` instance for `ToJSONFilter` before. I thought that it would straightforward at least to add a `[Text] -> a` instance without removing the existing `[String] -> a` instance, but I now realize that having both may create (admittedly obscure) ambiguous situations with `OverloadedStrings` that I do not think are present right now, so one instance should be omitted if that is to be avoided. I would prefer the `[Text] -> a` instance personally.

jgm:
I'm just thinking about the overall game plan.  There are a couple of options:

1.  Go in steps. First change only Text.Pandoc.Definition, and then make the necessary updates in pandoc, texmath, and pandoc-citeproc.  If Builder still used strings, then in this phase most of the readers would need few changes, since they use Builder to construct the AST.  Once that is all working, Builder could be switched to use Text, and then we'd have a round of further changes mostly affecting the readers.

2.  Change everything in pandoc-types now, and use that to show us what needs changing in the rest of the pandoc ecosystem.  With this approach, everything would change in one big step.

@despresc - are you planning to move on to pandoc and the other parts of the ecosystem after these changes?  Is this transition one you think could be done fairly quickly, or is this the beginning of a long haul?  The answers to these questions might affect a choice between 1 and 2.

jgm:
PS. We've already dropped support for GHC 7.10 in pandoc, so I have no worries about doing that here.  Indeed, I'll push a change to CI now.

jgm:
I've switched to using GitHub Actions CI instead of travis + appveyor, and testing only on ghc 8+.  You may want to rebase against master.

despresc:
I had planned to modify `texmath`, `pandoc`, and `pandoc-citeproc` in their entirety (in that order, I think, given their dependencies) if this pull request was successful. I do not think this will take an extremely long time, but there are some parts of each package I am not totally familiar with, so there is a chance for unusual difficulties to surface.

Most of the complexity is in `pandoc`. The changes required in the code are fairly straightforward (using `Data.Text` analogues of `String` functions, having the readers use a `Text` stream, redefining types to use `Text`, and so on), and most of the many `pandoc` modules are relatively unconnected to each other, which will help the internal changes go more smoothly. Certain changes, particularly to widely-used modules like `Text.Pandoc.Shared`, will require touching most of the modules at once. These will require large commits and will have to be done in one go, but the changes themselves should not be hard.

I am not as familiar with `texmath` and `pandoc-citeproc` as I am with `pandoc`, but nothing seemed overly difficult in those packages.

jgm:
That's great!  I think you're right, the changes are mostly very straightforward -- it's just that there are a lot of them.  I take it, then, that you'd prefer approach (2) - modifying Builder now and just changing everything to use Text?

despresc:
I prefer approach (2). That approach would minimize the number of widespread changes that other developers working on those packages would have to adapt to, since the changes are very non-local, and it should simplify my task. It assumes that the changes could be made fairly quickly, but I think that is possible. At the very least, the modification of `texmath` will provide some indication of how long the other two packages will take.

despresc:
I modified a local copy of `texmath` to use the most recent commit in the pull request, and changing the `Text.Pandoc.Definition` imports to `Text.Pandoc.Legacy.Definition` in the modules that had them was all that was required to allow `texmath` to compile and succeed in its tests. That bodes well for the initial propagation of the `pandoc-types` changes, at least.

jgm:
Okay, that makes sense.  (I didn't expect texmath to be difficult, because there's just one module that touches pandoc definitions.  pandoc and pandoc-citeproc will be more work.)

jgm:
I have manually merged this and created a new `text` branch.
I think that for now it should not be in `master`; when the pandoc and pandoc-citeproc changes are done we can merge it into master.

Note: there was a very stale old `text` branch that I deleted before pushing the new one.

I'm delighted you're taking this on!  This has been needed for a long time.

jgm:
I'm going to keep this issue open until we merge into `master` (after pandoc and pandoc-citeproc changes are complete).

jgm:
Released 1.20 with these changes!

Issue #65, opened 2020-03-21
Improving tables


despresc:
See the [main todo list](https://github.com/jgm/pandoc/issues/5581) and the [relevant issue](https://github.com/jgm/pandoc/issues/1024). I would like to start implementing better table handling in Pandoc. Specifically, I would implement all but the last of [these bullet points](https://github.com/jgm/pandoc/issues/1024#issuecomment-302919730) using one of the designs below (or a modified version of one of them).

I think something like [this recently outlined approach](https://github.com/jgm/pandoc/issues/1024#issuecomment-502589674) is a good way forward for now. The representation is a little loose (any table in the intermediate representation is valid, so there are multiple ways to write a given table, but only one normalized way), but it should allow the readers and writers to be switched more easily. This is slightly modified version of that approach:

```haskell
type RowSpan = Int
type ColSpan = Int
type Caption = [Block]
type ShortCaption = [Inline]
type ColWidth = Maybe Double
data CellType = DataCell | HeaderCell
data Cell = Cell Attr CellType (Maybe Aligment) RowSpan ColSpan [Block]
type Row = [Cell]

data Block =
  ...
  | Table Attr Caption ShortCaption [(Alignment, ColWidth)] [Row]
  ...
```

The `Maybe Alignment` on the individual cells allows the cells to override the alignment of the column(s) in which they reside. This makes it easier to specify one's intentions when a cell spans multiple columns with conflicting alignments, and has the advantage of allowing better `\multicolumn` and `\multirow` support in the LaTeX reader and writer. It also comes up naturally when one thinks of possible extensions to the supported markdown table formats.

A similar design has the following modifications:
```haskell
data Cell = Cell Attr (Maybe Alignment) RowSpan ColSpan [Block]
data HeaderRow = Row Attr [Cell]
data BodyRow = Row Attr [Cell] [Cell]

data Block =
  ...
  | Table Attr Caption ShortCaption [(Alignment, ColWidth)] [HeaderRow] [BodyRow] [HeaderRow]
  ...
```
This has the advantage of making explicit the table head/body/foot and row head/body structure that seems to be assumed in the first approach, where the first entirely header rows become the table head, and the last such rows become the table foot. Cells in the head and foot sections would correspond to `th` cells, and cells in body section would correspond to `td` cells. It does not require a `CellType`, but one could still be added, making these even more similar to HTML tables. This approach has the disadvantage of making the table representation more complex.

-----

[...]

PR #66, opened 2020-03-26
Better tables

despresc:
The initial implementation of issue #65. Other differences:

- The `Builder.table` and `Builder.simpleTable` have been adapted to the new type but have not had their signature or behaviour changed.
- The `Arbitrary` behaviour for tables is now different. Most notably, the instances do not produce normalized tables.
- The `Legacy.Definition` and `Legacy.Builder` modules are now not exposed. Would you like me to update them in some way, or remove them?
- There are `ToJSON` and `FromJSON` instances for `Row`, `Cell`, and `Caption`, all tagged. I checked the JSON output of the serialization test case and it seemed fine, so I updated the [expected bytestring](https://github.com/jgm/pandoc-types/compare/master...despresc:better-tables#diff-01f4ffe52cf097ab9ff89eebce394c86R369) in the test.
- `Walkable` instances for the new types have been added.

Once `pandoc` has been adapted to these changes, I can update this pull request with modified `table` and `simpleTable` builders. Incidentally, the behaviour of these (in particular `table`) should be pinned down. The `table` builder itself could be:
```haskell
table :: Caption
      -> [ColSpec]
      -> Rows
      -> Rows
      -> Rows
      -> Blocks
```
with `type Rows = Many Row` and a `row` and `cell` builder. Should the `table` builder normalize its output? If so, then I can put a table normalization function in `Builder` instead of somewhere in `pandoc`.

### DIFF
--- a/Text/Pandoc/Definition.hs
+++ b/Text/Pandoc/Definition.hs
@@ -64,18 +64,19 @@ module Text.Pandoc.Definition ( Pandoc(..)
                               , Attr
                               , nullAttr
                               , Caption(..)
-                              , RowHeadColumns
+                              , ShortCaption
+                              , RowHeadColumns(..)
                               , Alignment(..)
-                              , ColWidth
+                              , ColWidth(..)
                               , ColSpec
                               , Row(..)
-                              , TableHead
-                              , TableBody
-                              , TableFoot
+                              , TableHead(..)
+                              , TableBody(..)
+                              , TableFoot(..)
                               , Cell(..)
                               , emptyCell
-                              , RowSpan
-                              , ColSpan
+                              , RowSpan(..)
+                              , ColSpan(..)
                               , QuoteType(..)
                               , Target
                               , MathType(..)
@@ -211,9 +212,11 @@ instance Eq Format where
 instance Ord Format where
   compare (Format x) (Format y) = compare (T.toCaseFold x) (T.toCaseFold y)
 
--- | The width of the row head of each row of a table. The row body
--- takes up the remaining table width.
-type RowHeadColumns = Int
+-- | The number of columns taken up by the row head of each row of a
+-- 'TableBody'. The row body takes up the remaining columns.
+newtype RowHeadColumns = RowHeadColumns
+  { getRowHeadColumns :: Int
+  } deriving (Eq, Ord, Show, Read, Typeable, Data, Generic, Num, Enum)
 
 -- | Alignment of a table column.
 data Alignment = AlignLeft
@@ -221,45 +224,54 @@ data Alignment = AlignLeft
                | AlignCenter
                | AlignDefault deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
 
--- | The relative width of a column. A @Nothing@ value represents the
--- default width.
-type ColWidth = Maybe Double
+-- | The relative width of a column.
+data ColWidth = ColWidth Double
+              | ColWidthDefault deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
 
 -- | The specification for a single table column.
-type ColSpec = (Alignment, Maybe Double)
+type ColSpec = (Alignment, ColWidth)
 
--- | A table row. The header cells of the row are determined by the
--- value of the 'RowHeadColumns' of the table and the ultimate
--- placement of the table cells on the grid.
+-- | A table row.
 data Row = Row Attr [Cell]
   deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
 
 -- | The head of a table.
-type TableHead = [Row]
+data TableHead = TableHead Attr [Row]
+  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
 
--- | The body of a table.
-type TableBody = [Row]
+-- | A body of a table, with an intermediate head and the specified
+-- number of row header columns.
+data TableBody = TableBody Attr RowHeadColumns [Row] [Row]
+  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
 
 -- | The foot of a table.
-type TableFoot = [Row]
+data TableFoot = TableFoot Attr [Row]
+  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
+
+-- | A short caption, for use in, for instance, lists of figures.
+type ShortCaption = [Inline]
 
 -- | The caption of a table, with an optional short caption.
-data Caption = Caption (Maybe [Inline]) [Block]
+data Caption = Caption (Maybe ShortCaption) [Block]
   deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
 
 -- | A table cell.
-data Cell = Cell Attr (Maybe Alignment) RowSpan ColSpan [Block]
+data Cell = Cell Attr Alignment RowSpan ColSpan [Block]
   deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
 
 -- | A 1×1 empty cell.
 emptyCell :: Cell
-emptyCell = Cell nullAttr Nothing 1 1 []
+emptyCell = Cell nullAttr AlignDefault 1 1 []
 
 -- | The number of rows occupied by a cell; the height of a cell.
-type RowSpan = Int
+newtype RowSpan = RowSpan
+  { getRowSpan :: Int
+  } deriving (Eq, Ord, Show, Read, Typeable, Data, Generic, Num, Enum)
 
 -- | The number of columns occupied by a cell; the width of a cell.
-type ColSpan = Int
+newtype ColSpan = ColSpan
+  { getColSpan :: Int
+  } deriving (Eq, Ord, Show, Read, Typeable, Data, Generic, Num, Enum)
 
 -- | Block element.
 data Block
@@ -290,8 +302,8 @@ data Block
     | HorizontalRule
     -- | Table, with attributes, caption, optional short caption,
     -- column alignments and widths (required), table head, table
-    -- body, and table foot
-    | Table Attr Caption [ColSpec] RowHeadColumns TableHead TableBody TableFoot
+    -- bodies, and table foot
+    | Table Attr Caption [ColSpec] TableHead [TableBody] TableFoot
     -- | Generic block container with attributes
     | Div Attr [Block]
     -- | Nothing
@@ -512,6 +524,18 @@ instance ToJSON Alignment where
             AlignCenter  -> "AlignCenter"
             AlignDefault -> "AlignDefault"
 
+instance FromJSON ColWidth where
+  parseJSON (Object v) = do
+    t <- v .: "t" :: Aeson.Parser Value
+    case t of
+      "ColWidth"        -> ColWidth <$> v .: "c"
+      "ColWidthDefault" -> return ColWidthDefault
+      _     -> mempty
+  parseJSON _ = mempty
+instance ToJSON ColWidth where
+  toJSON (ColWidth ils)  = tagged "ColWidth" ils
+  toJSON ColWidthDefault = taggedNoContent "ColWidthDefault"
+
 instance FromJSON Row where
   parseJSON (Object v) = do
     t <- v .: "t" :: Aeson.Parser Value
@@ -533,6 +557,69 @@ instance FromJSON Caption where
   parseJSON _ = mempty
 instance ToJSON Caption where
   toJSON (Caption mshort body) = tagged "Caption" (mshort, body)
+
+instance FromJSON RowSpan where
+  parseJSON (Object v) = do
+    t <- v .: "t" :: Aeson.Parser Value
+    case t of
+      "RowSpan" -> RowSpan <$> v .: "c"
+      _         -> mempty
+  parseJSON _ = mempty
+instance ToJSON RowSpan where
+  toJSON (RowSpan h)  = tagged "RowSpan" h
+
+instance FromJSON ColSpan where
+  parseJSON (Object v) = do
+    t <- v .: "t" :: Aeson.Parser Value
+    case t of
+      "ColSpan" -> ColSpan <$> v .: "c"
+      _         -> mempty
+  parseJSON _ = mempty
+instance ToJSON ColSpan where
+  toJSON (ColSpan w)  = tagged "ColSpan" w
+
+instance FromJSON RowHeadColumns where
+  parseJSON (Object v) = do
+    t <- v .: "t" :: Aeson.Parser Value
+    case t of
+      "RowHeadColumns" -> RowHeadColumns <$> v .: "c"
+      _                -> mempty
+  parseJSON _ = mempty
+instance ToJSON RowHeadColumns where
+  toJSON (RowHeadColumns w)  = tagged "RowHeadColumns" w
+
+instance FromJSON TableHead where
+  parseJSON (Object v) = do
+    t <- v .: "t" :: Aeson.Parser Value
+    case t of
+      "TableHead" -> do (attr, body) <- v .: "c"
+                        return $ TableHead attr body
+      _           -> mempty
+  parseJSON _ = mempty
+instance ToJSON TableHead where
+  toJSON (TableHead attr body) = tagged "TableHead" (attr, body)
+
+instance FromJSON TableBody where
+  parseJSON (Object v) = do
+    t <- v .: "t" :: Aeson.Parser Value
+    case t of
+      "TableBody" -> do (attr, rhc, hd, body) <- v .: "c"
+                        return $ TableBody attr rhc hd body
+      _           -> mempty
+  parseJSON _ = mempty
+instance ToJSON TableBody where
+  toJSON (TableBody attr rhc hd body) = tagged "TableBody" (attr, rhc, hd, body)
+
+instance FromJSON TableFoot where
+  parseJSON (Object v) = do
+    t <- v .: "t" :: Aeson.Parser Value
+    case t of
+      "TableFoot" -> do (attr, body) <- v .: "c"
+                        return $ TableFoot attr body
+      _           -> mempty
+  parseJSON _ = mempty
+instance ToJSON TableFoot where
+  toJSON (TableFoot attr body) = tagged "TableFoot" (attr, body)
 
 instance FromJSON Cell where
   parseJSON (Object v) = do
@@ -619,8 +706,8 @@ instance FromJSON Block where
       "Header"         -> do (n, attr, ils) <- v .: "c"
                              return $ Header n attr ils
       "HorizontalRule" -> return HorizontalRule
-      "Table"          -> do (attr, cpt, align, rhw, hdr, body, foot) <- v .: "c"
-                             return $ Table attr cpt align rhw hdr body foot
+      "Table"          -> do (attr, cpt, align, hdr, body, foot) <- v .: "c"
+                             return $ Table attr cpt align hdr body foot
       "Div"            -> do (attr, blks) <- v .: "c"
                              return $ Div attr blks
       "Null"           -> return Null
@@ -638,8 +725,8 @@ instance ToJSON Block where
   toJSON (DefinitionList defs) = tagged "DefinitionList" defs
   toJSON (Header n attr ils) = tagged "Header" (n, attr, ils)
   toJSON HorizontalRule = taggedNoContent "HorizontalRule"
-  toJSON (Table attr caption aligns rhw hd body foot) =
-    tagged "Table" (attr, caption, aligns, rhw, hd, body, foot)
+  toJSON (Table attr caption aligns hd body foot) =
+    tagged "Table" (attr, caption, aligns, hd, body, foot)
   toJSON (Div attr blks) = tagged "Div" (attr, blks)
   toJSON Null = taggedNoContent "Null"
 
@@ -673,8 +760,13 @@ instance NFData MetaValue
 instance NFData Meta
 instance NFData Citation
 instance NFData Alignment
+instance NFData RowSpan
+instance NFData ColSpan
 instance NFData Cell
 instance NFData Row
+instance NFData TableHead
+instance NFData TableBody
+instance NFData TableFoot
 instance NFData Caption
 instance NFData Inline
 instance NFData MathType
@@ -683,6 +775,8 @@ instance NFData CitationMode
 instance NFData QuoteType
 instance NFData ListNumberDelim
 instance NFData ListNumberStyle
+instance NFData ColWidth
+instance NFData RowHeadColumns
 instance NFData Block
 instance NFData Pandoc
 
--- a/Text/Pandoc/Definition.hs
+++ b/Text/Pandoc/Definition.hs
@@ -66,6 +66,7 @@ module Text.Pandoc.Definition ( Pandoc(..)
                               , Caption(..)
                               , Alignment(..)
                               , ColWidth
+                              , ColSpec
                               , Row(..)
                               , RowHead
                               , RowBody
@@ -221,6 +222,9 @@ data Alignment = AlignLeft
 -- default width.
 type ColWidth = Maybe Double
 
+-- | The specification for a single table column.
+type ColSpec = (Alignment, Maybe Double)
+
 -- | A table row
 data Row = Row Attr RowHead RowBody
   deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
@@ -291,7 +295,7 @@ data Block
     -- | Table, with attributes, caption, optional short caption,
     -- column alignments and widths (required), table head, table
     -- body, and table foot
-    | Table Attr Caption [(Alignment, ColWidth)] TableHead TableBody TableFoot
+    | Table Attr Caption [ColSpec] TableHead TableBody TableFoot
     -- | Generic block container with attributes
     | Div Attr [Block]
     -- | Nothing
--- a/Text/Pandoc/Definition.hs
+++ b/Text/Pandoc/Definition.hs
@@ -74,7 +74,6 @@ module Text.Pandoc.Definition ( Pandoc(..)
                               , TableBody(..)
                               , TableFoot(..)
                               , Cell(..)
-                              , emptyCell
                               , RowSpan(..)
                               , ColSpan(..)
                               , QuoteType(..)
@@ -258,10 +257,6 @@ data Caption = Caption (Maybe ShortCaption) [Block]
 -- | A table cell.
 data Cell = Cell Attr Alignment RowSpan ColSpan [Block]
   deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
-
--- | A 1×1 empty cell.
-emptyCell :: Cell
-emptyCell = Cell nullAttr AlignDefault 1 1 []
 
 -- | The number of rows occupied by a cell; the height of a cell.
 newtype RowSpan = RowSpan Int
--- a/Text/Pandoc/Definition.hs
+++ b/Text/Pandoc/Definition.hs
@@ -57,14 +57,25 @@ module Text.Pandoc.Definition ( Pandoc(..)
                               , docDate
                               , Block(..)
                               , Inline(..)
-                              , Alignment(..)
                               , ListAttributes
                               , ListNumberStyle(..)
                               , ListNumberDelim(..)
                               , Format(..)
                               , Attr
                               , nullAttr
-                              , TableCell
+                              , Caption(..)
+                              , Alignment(..)
+                              , ColWidth
+                              , Row(..)
+                              , RowHead
+                              , RowBody
+                              , TableHead
+                              , TableBody
+                              , TableFoot
+                              , Cell(..)
+                              , emptyCell
+                              , RowSpan
+                              , ColSpan
                               , QuoteType(..)
                               , Target
                               , MathType(..)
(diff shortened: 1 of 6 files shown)


Text/Pandoc/Definition.hs
```diff
@@ -209,31 +211,91 @@ instance Eq Format where
...
+    | HorizontalRule
+    -- | Table, with attributes, caption, optional short caption,
+    -- column alignments and widths (required), table head, table
+    -- body, and table foot
+    | Table Attr Caption [(Alignment, ColWidth)] TableHead TableBody TableFoot
+    -- | Generic block container with attributes
+    | Div Attr [Block]
+    -- | Nothing
+    | Null
     deriving (Eq, Ord, Read, Show, Typeable, Data, Generic)
```
jgm: Can you really do the haddocks this way? I didn't know this form could be used for constructors.

jgm:
> with type Rows = Many Row

I would just use `[Row]`.  I only use the `Many` for `Inlines` and `Blocks` (mainly because we need a sequence there for efficient concatenation -- I don't think that applies to the rows).

> Should the table builder normalize its output? If so, then I can put a table normalization function in Builder instead of somewhere in pandoc.

This probably makes sense, but what exactly did you have in mind for normalization?

Text/Pandoc/Definition.hs
despresc: It's documented [here](https://haskell-haddock.readthedocs.io/en/latest/markup.html#constructors-and-record-fields). The generated haddocks seem to be fine, in any case. I changed it because the documentation source for `Block` was getting a little hard to read in the other style.

jgm: Yeah, the `|` in the haddocks is too easily confused with the `|` in the type definition.

despresc:
I talked about table normalization a bit at the end of my [first comment](https://github.com/jgm/pandoc-types/issues/65#issue-585562115). Looking at [the diagram I drew](https://github.com/jgm/pandoc-types/issues/65#issuecomment-602652104), a table would be normalized by applying that procedure to each of its six sections, with the widths and heights of the grids involved being determined by the geometry of the initial table (number of rows in the `TableHead`, `TableBody`, and `TableFoot` for heights, number of cells in the first non-empty row head and the length of the `[ColSpec]` list for widths).

Briefly, when you lay out a normalized table on a grid of the correct dimensions, it will have no overlapping cells, no cells that cross section boundaries or the grid boundaries, no empty grid spaces, and will have individual sections with the appropriate dimensions so they fit against each other properly. For an arbitrary table, cell dimensions may be reduced and rows may be padded or truncated to produce such a normalized table, in a way that I think is consistent with HTML behaviour.

Text/Pandoc/Definition.hs
despresc: Would you like me to revert the change?

jgm:
I think it makes sense to add normalization in the builders.

Here's a thought, though (about a revision to the table type).

Instead of having each row consist of a list of row head cells, followed by a list of regular row cells, why not have a global parameter for the table that is just an integer, specifying the number of cells in each row that will be treated as row heads?

I suggest this because I'm thinking there are some formats that can't really represent tables with different numbers of row heads in each head.  That's easy to represent in HTML, of course. But in LaTeX, the way you'd represent a row head would be to insert a vertical line, and that has to be done in the table header.  So you wouldn't be able to handle a table with different numbers of row heads in each row.

Is there any practical application for tables with different numbers of row heads in each row? If not, maybe we should make these impossible to represent in the type, in the way I suggested.

despresc:
I can't think of any applications for varying row head widths offhand, in the current design. Single tables generally want to have their row heads be a consistent width. There might be examples of multi-part HTML tables that have varying row head sizes, but those can't be represented very well with the current type anyway, since those kinds of tables usually have intermediate `th` header rows and we also only allow one table body.

Some, but less, work would be required in normalization, since cells in the row head shouldn't go past the given width. The `Row` type would be simpler, at the (perhaps small) cost of not being able to tell immediately what cells are header cells in a row.

jgm:
It seems to me there's some good reason to go with this modification, then. What do you think?

despresc:
It sounds like a good change. I can update the pull request with something like a `type RowHeadWidth` parameter.

jgm:
sounds good.

jgm:
I suppose `Width` is somewhat ambiguous; someone might expect it to be a dimension rather than a number of cells.  Maybe `RowHeadCells`?

despresc:
I think it may need to be a width, to allow for row headers with multiple levels, like those in the "Data about the planets" table on [this MDN page](https://developer.mozilla.org/en-US/docs/Learn/HTML/Tables/Basics)

jgm:
Making it a width would be awkward though, since we allow "simple tables" with no explicit widths.  A simple table can still have row heads.  Isn't what we want a number of cells?  In the table you mention, it would be two -- but most of the rows would have a colspan'd cell that takes up the space of two cells in first position.  Maybe that's what you meant by "width"?  I was thinking of something like "2in" or "0.10" (a fractional width like our widths in ColSpec).

despresc:
I understand now. I think we agree that it should represent the total number of columns occupied by the row head, so `type RowHeadSomething = Int`. The name `RowHeadCells` suggested to me that a particular number of `Cell` values formed the row head in each row.

jgm:
`RowHeadSpan` seems as good as anything to me.

jgm:
I think this looks good -- but of course we may see otherwise when trying to implement the readers and writers.  Probably the best approach is to have your branch of pandoc depend on this tag from the repository.  I'd prefer not to merge this into pandoc-types until the pandoc changes have been done and we can see how everything works together.

Text/Pandoc/Definition.hs
jgm: I'm okay with this style, I guess both approaches have their drawbacks.
In a source code editor syntax highlighting will help avoid confusion, and in the generated haddocks it will be the same either awy.

despresc:
Yes, this shouldn't be merged yet. I believe that when the `Text` switch happened I initially used my own `pandoc-types` fork in `pandoc`, which was then merged into a branch in this repo (that I then used).

jgm:
I'm having second thoughts about `RowHeadSpan`.  I think a better name for this type would be `RowHeadColumns`.  Reason:  `RowHeadSpan` evokes "row span," a single cell spanning several rows.  We already have a notion of "columns" (e.g. `ColSpec`) and here we are designating how many of these columns are row heads (of course, in some cases you may have a cell that spans multiple columns, so this is a better name than `RowHeadCells`).

mb21:
> I'm having second thoughts about `RowHeadSpan`. I think a better name for this type would be `RowHeadColumns`.

Ha, that's funny... reading the source I was just going to propose `RowHeadSpan` (in order to be consistent with `RowSpan` and `ColSpan`. Is `RowHeadColSpan` too verbose?

Text/Pandoc/Definition.hs
```diff
@@ -209,31 +211,91 @@ instance Eq Format where
...
+type TableHead = [Row]
+
+-- | The body of a table.
+type TableBody = [Row]
+
+-- | The foot of a table.
+type TableFoot = [Row]
+
+-- | The caption of a table, with an optional short caption.
+data Caption = Caption (Maybe [Inline]) [Block]
```
mb21: are we thinking of potentially reusing this for [figure captions](https://github.com/jgm/pandoc/issues/3177#issuecomment-421639264), or...?

mb21:
@jgm I see we've given up on "secondary header rows" (or subheaders) in the middle of a table? I'm fine with that... just a question.

P.S. really great to see work on this getting seriously underway, thanks @despresc !

Text/Pandoc/Definition.hs
jgm: I think we should certainly re-use this for figure captions.  It wouldn't make sense for headings.

mb21:
Thanks for your replies @jgm... just making sure you saw these here as well... (about `RowHeadColSpan` and multi-level`)

jgm:
You're right that the current design doesn't allow for arbitrary header rows within the table.
Maybe that is something we should reconsider.  I don't see a problem with rendering these in HTML or LaTeX (not sure about other formats).  I guess it's partly a question of whether there's a real need for them.

Also @despresc has already put in a lot of work with the current design; not sure how hard it would be at this point to backtrack on this.

despresc:
It isn't hard to change right now. I'll have to redo the `pandoc` tests when I add the `ColWidth` data type and the various dimension newtypes anyway. I don't know if there is enough demand for such tables either, so I can't comment on whether the added expressiveness is worth it or not.

----

I can think of two main ways to have intermediate header rows, with many variations.

We could have
```haskell
data Row = HeaderRow Attr [Cell]
         | BodyRow Attr [Cell] [Cell]
data Block
  = ...
  | Table Attr Caption [ColSpec] [Row]
```
and have the first and last `HeaderRow` rows be the overall table head and foot. Intermediate spans of `HeaderRow` rows would be treated as intermediate header sections, and in normalization cells would not be allowed to cross the boundaries of the various sections. The row heads could also be constrained to occupy the same number of columns within each `BodyRow` section during normalization, if desired.

Or the `CellType` could be brought back, perhaps keeping the `TableHead`, `TableBody`, and `TableFoot`. I assume that the implicit structure would be roughly the same as the explicit structure above: rows entirely of `HeaderCell` cells would be header rows, and header rows would be grouped together into header sections (table head, intermediate header sections, table foot).

despresc:
Tables could have multi-part bodies, in line with the possibility of multiple `tbody` elements in HTML. Something like
```haskell
data Row = Row Attr [Cell]
data TableHead = TableHead Attr [Row]
data TableBody = TableBody Attr RowHeadColumns [Row] [Row] -- header, then body rows
data TableFoot = TableFoot Attr [Row]
data Block
  = ...
  | Table Attr Caption [ColSpec] TableHead [TableBody] TableFoot
```
which would still restrict the appearance of header cells to particular locations in the table, like the current new design does.

jgm:
or even
```
type HeadRows = Int
data TableBody = TableBody Attr HeadRows [Row]
```

despresc:
I don't think the `TableBody` having a single `[Row]` slot is that useful. It would still have to keep track of the row header cells, with `RowHeadColumns` or a modified `Row` type.

jgm:
OK, I think I'd be fine with the type you suggest.  @tarleb, @mb21, any feedback on this?

mb21:
Having both `TableHead` and `[Row] [Row] -- header, then body rows` seems a bit complicated? Also more trouble choosing which the canonical version is?

> it's not enough to take the first and last header row to be the table head and foot.

I'm not sure... I guess it could work? the question is whether this hack would be more trouble than the two different head representation proposed by @despresc.... I'm not sure?

Or something like this:

```hs
data Row = Row Attr [Cell]
data TableHead = TableHead Attr [Row]
data TableBody = TableBody Attr RowHeadColumns [Row]
data TableFoot = TableFoot Attr [Row]
data TablePart = TablePart TableHead TableBody TableFoot
data Block
  = ...
  | Table Attr Caption [ColSpec] [TablePart]
```

or to moving into [figure](https://github.com/jgm/pandoc/issues/3177) territory:

```hs
data Row = Row Attr [Cell]
data TableHead = TableHead Attr [Row]
data TableBody = TableBody Attr RowHeadColumns [Row]
data TableFoot = TableFoot Attr [Row]
data TablePart = TablePart [ColSpec] TableHead TableBody TableFoot
data Block
  = ...
  | Figure Attr Caption [TablePart]
```

despresc:
I think the table as a whole is supposed to have a distinguished head and foot, so the `TableHead` and `TableFoot` parts of the `Table` constructor would have to stay. That is what @jgm's earlier example suggested, since an intermediate header appears immediately after the table head. I don't have any objection to table body sections having a footer as well.

Actually, I am also interested in adding a `Figure` block (and a `label`/`ref` mechanism for that matter, though I haven't thought it through fully), but my impression of that issue was that something like `Figure Attr Caption [Block]`, for nested figures and greater flexibility, was more appealing to people. It might be better to keep the individual upgrades separate, though.

jgm:
> I think the table as a whole is supposed to have a distinguished head and foot, so the TableHead and TableFoot parts of the Table constructor would have to stay. 

Correct.

> I don't have any objection to table body sections having a footer as well.

I guess we need to figure out whether there's any real use for this, and how it would be represented in output formats.  (In HTML, for example, it's clear what an intermediate header row would be -- a tr with th elements -- but not so clear what an intermediate foot would look like.)

despresc:
For an HTML table with no `tbody` elements there would be no way of distinguishing an intermediate head from an intermediate foot. I suppose that `th` rows would all count as intermediate heads and the feet would be empty, for lack of a clearer way of interpreting that kind of table.

if the table does have `tbody` elements, then an intermediate foot is just the final span of `th` rows in that element.

despresc:
The writer would map our `TableBody` onto a `tbody` regardless, as I believe the current HTML writer does in spirit.

This interpretation does mean that intermediate header rows within `tbody` elements would not be recognized; the initial and final span of `th` rows would be the head and foot of the `TableBody` and any `th` cells outside of that or the `RowHeadColumns` would become data cells, in effect.

jgm:
I'm leaning towards the view that we shouldn't the additional complexity for intermediate footers unless there's a strong reason we need them.

tarleb:
The only comment I have is about each`Cell` including an `Alignment`: I understand that it can serve as a tie breaker for colspans, but I'm not fully convinced that the added complexity is worth it. We may have to iterate on the design anyway, so maybe we can start out with the simpler version?

All my other feedback would be pure bike-shedding, I'll keep that to myself.

despresc:
Multi-column cells often have alignments that directly contradict the alignments of the columns they span. Having the columns vote in some way on the alignment of these cells will tend to produce incorrect results, I think. Letting cells override their columns' alignments is the simplest way of avoiding this. The alignment parameter on, say, `\multicolumn` exists for this reason.

The other thing I was thinking about is extensions to `simple_tables`, so that you could write a table like this:
```
 head1    head2
-------  -------
 a   b   c     d
--- ---  --- ---
1    2    3   4
5    6    7   8
```
(The column alignments would be determined by the lowest header row, I suppose). Users would expect the header alignment rules to apply to every header.

Granted, this all refers to currently unimplemented features, but my feeling is that the need for individual cell alignments will be felt fairly quickly.

jgm:
Just so everyone can comment more easily, I've extracted the latest version of the table type changes here.  This looks good to me, I think, but it would be good for everyone to pass eyes over it now before pandoc changes are made:

``` haskell
-- | The number of columns taken up by the row head of each row of a
-- 'TableBody'. The row body takes up the remaining columns.
newtype RowHeadColumns = RowHeadColumns
  { getRowHeadColumns :: Int
  } deriving (Eq, Ord, Show, Read, Typeable, Data, Generic, Num, Enum)

-- | Alignment of a table column.
data Alignment = AlignLeft
               | AlignRight
               | AlignCenter
               | AlignDefault deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)

-- | The relative width of a column.
data ColWidth = ColWidth Double
              | ColWidthDefault deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)

-- | The specification for a single table column.
type ColSpec = (Alignment, ColWidth)

-- | A table row.
data Row = Row Attr [Cell]
  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)

-- | The head of a table.
data TableHead = TableHead Attr [Row]
  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)

-- | A body of a table, with an intermediate head and the specified
-- number of row header columns.
data TableBody = TableBody Attr RowHeadColumns [Row] [Row]
  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)

-- | The foot of a table.
data TableFoot = TableFoot Attr [Row]
  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)

-- | A short caption, for use in, for instance, lists of figures.
type ShortCaption = [Inline]

-- | The caption of a table, with an optional short caption.
data Caption = Caption (Maybe ShortCaption) [Block]
  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)

-- | A table cell.
data Cell = Cell Attr Alignment RowSpan ColSpan [Block]
  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)

-- | A 1×1 empty cell.
emptyCell :: Cell
emptyCell = Cell nullAttr AlignDefault 1 1 []

-- | The number of rows occupied by a cell; the height of a cell.
newtype RowSpan = RowSpan
  { getRowSpan :: Int
  } deriving (Eq, Ord, Show, Read, Typeable, Data, Generic, Num, Enum)

-- | The number of columns occupied by a cell; the width of a cell.
newtype ColSpan = ColSpan
  { getColSpan :: Int
  } deriving (Eq, Ord, Show, Read, Typeable, Data, Generic, Num, Enum)

data Block = ....
    -- | Table, with attributes, caption, optional short caption,
    -- column alignments and widths (required), table head, table
    -- bodies, and table foot
    | Table Attr Caption [ColSpec] TableHead [TableBody] TableFoot
```

jgm:
I noticed this issue:
https://github.com/jgm/pandoc/issues/5998
While we're at it, should we add a constructor for `AlignDecimal`?

despresc:
I think LaTeX has the [dcolumn package](https://www.ctan.org/pkg/dcolumn), and there is an experimental `text-align: string` property in a [draft of CSS4](https://drafts.csswg.org/css-text-4/#alignment) (unimplemented by every browser, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/text-align#Browser_compatibility)). The common feature of both (and apparently MultiMarkDown) is that alignment happens on any specified character. Perhaps it should be `AlignString Text`? If it is to be added.

There is also [siunitx](https://www.ctan.org/pkg/siunitx), which includes an `S` column alignment for alignment on a decimal marker, so that could be used if `AlignDecimal` were the choice. I suppose cells would get an `aligndecimal` class in HTML in that scenario?

I have no idea about the other formats.

despresc:
It would be fairly straightforward to add later. Since it's only adding a constructor, the existing tests would not have to change immediately, the readers would not be affected, and the writers would just need to convert it to `AlignDefault` in the short term.

It's also not much extra work for me to change it right now, either.

jgm:
Tempted to leave out decimal alignment for now...it introduces a number of additional issues which maybe it would be better to avoid for now?

despresc:
I agree. It might need more thought. It would certainly be nice to have in future, even if only a limited number of formats support it.

despresc:
It depends on exactly how we want the cell alignment to behave. The cell alignment was originally `Maybe Alignment` so that `Just align` would be an explicit override and `Nothing` would result in some alignment being calculated (perhaps just `AlignDefault` unconditionally, or some kind of vote from the columns). The current iteration has `AlignDefault` representing no override, I believe.

I think that having any cell be capable of overriding the column alignment is simpler conceptually, so that `RowSpan` and `ColSpan` only need to represent the dimensions of the cell. Maybe the cell alignment could be `Maybe Alignment` or `CellAlign = CellAlign Alignment | CellAlignDefault` again, if that's clearer?

If it is important that cells that don't span columns not be able to override the column alignments, then your type is better.

mb21:
Looks good to me, and at some point we have to stop the bike-shedding :-)

nit-pick: would add to the doc string to something like:

    -- | The relative width of a column, ranging 0-1.

tarleb:
> If it is important that cells that don't span columns not be able to override the column alignments, then your type is better.

I don't think it's important, either way should work well, really. I think I just like the restrictiveness of pandoc's AST, and the more HTML-ish power takes some getting used to.

jgm:
One thing that hasn't come up is alignment for row spans (and vertical alignment generally).  We don't currently have any way of representing that (other than the generic Attr).  But I do worry a bit about making the type *too* expressive; as @tarleb notes, the restrictiveness of pandoc's types can be an advantage in some ways (making it easier to write readers and writers).

despresc:
Cells could get a `data AlignVert = AlignTop | AlignMid | AlignBot`, perhaps with a `data CellAlignment` that bundles the horizontal and vertical alignments together. That is something that tends to appear when `RowSpan` cells are allowed, but I don't know about reader and writer support for it.

For now, I've adapted `pandoc` to the current changes in this package. The last thing to do is change the `Builder` module as we discussed. Changes to the `Table`-related types are still relatively easy for me to implement now, before `Builder` is changed and those changes get propagated to `pandoc`.

despresc:
I've realized while writing some tests for the new `Builder.table` that it's actually somewhat difficult to determine what cells are row head cells in a `TableBody`. You might think that if you have a function `cellWidth (Cell _ _ _ (ColSpan w) _) = w`, then taking cells just until their total `cellWidth` meets or exceeds `RowHeadColumns` would be enough, but that only works for the very first row in a `TableBody`. Cells in earlier rows may descend into later rows, occupying grid positions in those rows and so decreasing their effective `RowHeadColumns`.

So to determine which cells are row head cells in a particular row, you need to have kept track of how far down the cells of all the previous rows descend beyond their rows. This isn't a problem if you're operating on an entire table, say to render it, since you will need to do that anyway while checking (and correcting) the well-formedness of the table. But it seems excessively annoying to someone who, for instance, wants to write a filter that only adjusts the data cells of a table, or really anyone who needs to operate on only parts of a table in isolation from the table as a whole.

This problem could be fixed by bringing back `data HeaderRow = HeaderRow Attr [Cell]` and `data BodyRow = BodyRow Attr [Cell] [Cell]`, or changing `Row` to `Row Attr [Cell] [Cell]`. The `RowHeadColumns` could be kept and used to constrain the row head section of a `Row`, or it could be removed and have it be inferred from the rows of a `TableBody`.

jgm:
It seems to me that the problem of inferring `RowHeadColumns` from the table body (bodies) is going to face exactly the same difficulty you've described (in case there are row spans).  Since we'll need `RowHeadColumns` for the whole table in some formats (e.g. LaTeX where we'll want a vertical line, which needs to be specified in the table header), if we require that this be inferred, we'll be facing this problem anyway and won't avoid it.  On the other hand, if we keep it *and* distinguish row heads from other cells in each Row, then our type will be able to represent inconsistent tables.  Given those two alternatives I wonder whether the present situation is so bad?

despresc:
If `RowHeadColumns` were kept around, it would become an explicit number of columns that the row head would have to occupy. I don't think that's too different from having the table width being set to the length of the `[ColSpec]` of the table.

If it weren't kept, then the total `cellWidth` of the first non-empty row head in a `TableBody` could become its `RowHeadColumns`, since descending cells wouldn't be a problem in that case: row body cells always start to the right of any row head cell on the grid. But maybe that's too implicit.

This is mostly for the benefit of non-`pandoc` users of these types, since I'm not aware of table processing in `pandoc` that doesn't involve operating on the entire table.

jgm:
> If it weren't kept, then the total cellWidth of the first non-empty row head in a TableBody could become its RowHeadColumns, since descending cells wouldn't be a problem in that case: row body cells always start to the right of any row head cell on the grid. But maybe that's too implicit.

I guess that could work -- unless the table doesn't have a body (rare but possible).

> This is mostly for the benefit of non-pandoc users of these types, since I'm not aware of table processing in pandoc that doesn't involve operating on the entire table.

I guess I'm a little reluctant to make the type more complex just to support a hypothetical non-pandoc use case.  But maybe it's okay.  I do see how a certain kind of thing could be more difficult without this.

despresc:
Certainly, it is hypothetical. I can leave it be, since in some sense it's just caching data about the table bodies. A function could always be exposed (say in `pandoc`) that unzips a `TableBody` into its row head and row body, if there were great demand for it.

despresc:
Incidentally, when you say

> I guess that could work -- unless the table doesn't have a body (rare but possible).

do you mean the body of a `TableBody`? Right now the `RowHeadColumns` are stored in each `TableBody` so that they can vary between them.

jgm:
Ah right.  I'd forgotten that.
```
-- | A body of a table, with an intermediate head and the specified
-- number of row header columns.
data TableBody = TableBody Attr RowHeadColumns [Row] [Row]
```
Can you remind me why we didn't do
```
newtype HeadRows = HeadRows Int
data TableBody = TableBody Attr HeadRows RowHeadColumns [Row]
```

despresc:
To me, the body is ultimately a pair of lists of `Row`. It seemed odd to represent `([Row], [Row])` as `(Int, [Row])` when the length of the first list isn't unusually important, overall or in comparison to the length of the second list. But if we're already not distinguishing between header and data cells very much, maybe the second is more consistent in your view?

jgm:
I guess I was thinking it looked a bit inconsistent to use a number to indicate the header row cells but two lists to indicate the header rows.  Not a huge issue, but it did raise a question.

In processing these, if we end up doing `take n rows, drop n rows`, then we might as well have had two lists to begin with.  But I suppose another option would be something like
``` haskell
zipWith toRow (replicate headerRows True ++ repeat False) row
```
where `toRow` takes an initial boolean parameter indicating whether it's a header row.

despresc:
I've done a preliminary adaptation of `Builder.table` to the new type, and added more builder functions. The `simpleTable` builder still has its old signature, since I wasn't sure exactly how much more complex it should become, if at all.

despresc:
Two examples:

```haskell
-- Unnormalized TableHead with cells represented as (label, RowSpan, ColSpan)
-- and a target table width of 3

[[(a, 1, 1),(b, 2, 3)]
,[(c, 2, 2),         ,(d, 1, 1)]]
```
becomes
```
+---+---+---+     [[(a, 1, 1),(b, 2, 2)]
| a |       |     ,[(c, 1, 1)]         ]
+---+   b   +
| c |       |
+---+---+---+
```
and
```haskell
-- Unnormalized TableBody with intermediate header, with a
-- target width of 3 and a RowHeadColumns of 1.

[[(a, 2, 2),(b, 1, 1)]]

[[(c, 3, 2),(d, 2, 1),(e, 2, 1)]
,[]]
```
becomes
```
-- '#' used to separate the row head from the row body

+---+---+---+   
| a # b | - |	[[(a, 1, 1),(b, 1, 1),emptyCell]]                                 
+===+===+===+	
| c # d | e |	[[(c, 2, 1),(d, 2, 1),(e, 2, 1)] 
+   +   +   +	,[]]                            
|   #   |   |
+---+---+---+


```

despresc:
It's actually somewhat difficult to ensure that the table is "tight", in the sense that no or column is ever totally empty, since correcting this in a particular row or column can require shrinking cells arbitrarily early in the table, potentially multiple times. This is why no empty row filtering is performed right now in `table` or the normalization functions. Note that `table` did not do that before this update, either.

ickc:
I don’t worry about empty row/column but I worry about the error handling when collisions as shown in the examples happened.

Is it going to handle it silently?

Eg can we construct a grid and count how many cells lies in those grids, and if the count is greater than 1 then emit a warning message telling user which cells are colliding?

mb21:
> It's actually somewhat difficult to ensure that the table is "tight", in the sense that no row or column is ever totally empty, since correcting this in a particular row or column can require shrinking cells arbitrarily early in the table, potentially multiple times.

I think HTML spec solves this by then expanding the table? I think they specced it in a way that browsers can do just one pass over the table... so earlier cells always "win" over later cells kind of.

despresc:
Right now it's silent, @ickc, partly because the error and logging functionality all live in `pandoc` and so can't be used here. I think it would require cooperation between the builder and the readers to generate useful warnings, since the readers don't pass along detailed source information. Maybe other people can comment on what sort of warning behaviour here would be consistent with the rest of pandoc. 

Looking over the HTML spec, @mb21, I think it's simply declared a table model error if empty grid rows or columns or overlapping cells occur, and user agents can choose how to render the result. It feels possible to eliminate both in one pass, but I haven't actually implemented it. I wasn't sure if there was sufficient demand for it to justify the effort, as it seemed a bit involved.

tarleb:
I think it's ok to follow the "garbage-in, garbage-out" philosophy here: if the input is not a well-formed table, then we should be free to produce bad output. Validation should be handled by a different library, as proposed by @Mercerenies.

Text/Pandoc/Definition.hs
```diff
@@ -209,31 +211,96 @@ instance Eq Format where
 instance Ord Format where
   compare (Format x) (Format y) = compare (T.toCaseFold x) (T.toCaseFold y)
 
+-- | The number of columns taken up by the row head of each row of a
+-- 'TableBody'. The row body takes up the remaining columns.
+newtype RowHeadColumns = RowHeadColumns Int
+  deriving (Eq, Ord, Show, Read, Typeable, Data, Generic, Num, Enum)
```
mb21, line 217: just stumbled over this tweet... https://twitter.com/mxlambda/status/1248228970694402048?s=12 :-)

stroobandt:
# The CALS table model
Perhaps of interest to this discussion is the [the CALS table model](https://en.wikipedia.org/wiki/CALS_Table_Model).
[CALS](https://en.wikipedia.org/wiki/CALS_%28DOD%29) started out as a United States Department of Defense (DoD) initiative for creating a DTD standard for representing tables in SGML/XML.

CALS allows for cell merging.

CALS is also the historic precursor of the HTML table model. A little known fact is that CALS is supported in LaTeX through the `cals` package. In fact, it is easier to use and offers more capabilities than the more commonly employed `tabularx` package in LaTeX.

Here is the documentation of the `cals` package in LaTeX:

 - <http://www.ctan.org/tex-archive/macros/latex/contrib/cals> (demo & doc)
 - <http://tug.org/TUGboat/tb32-2/tb101parashchenko.pdf> (user's manual)

jgm:
@stroobandt - Since the table type has already been implemented, a better place for your comment is probably jgm/pandoc#6313.  cals might indeed be a better fit to our internal model than standard tabular, tabularx, or longtable.  Note, however, that we've historically been averse to using less common packages in the default latex output.

mirh:
If row/column spans are now supported, shouldn't you update the documentation?

jgm:
@mirh If there is a particular place in the documentation that seems wrong, please point to the exact place or submit a PR. Note that row/column spans are still not supported in pandoc's markdown.

mirh:
Oh, I see, sorry then.

Issue #75, opened 2020-08-19
Single constructor data types and JSON serialization


boisgera:
Hi everyone !

Until recently I worked under the assumption that a pandoc-types datatype with a single constructor (say `Format`) had its type erased from the JSON representation : instead of `{"t": type, "c": content}`, the representation was simply `content`.

I think that (maybe with the exception of `Meta` ?) this assumption was valid until the recent changes to the document model. Now, AFAICT some types with a single constructor have their types erased and some don't. I thought for a moment that the difference was that some where declared with `newtype` keyword (type erasure) and some with `data` keyword (no type erasure) which  would make sense (if I understand correctly the difference between the two keywords in Haskell) but this second hypothesis doesn't hold either.

Could anyone explain me if there is a simple rule based on the definition of pandoc types that says if the type of the data will be erased in JSON representation ?

AFAICT, `Format` data (`newtype`) has its type erased in JSON, but `RowSpan` data (`newtype`) has its type serialized. `Cell` data (`data`) also have their types serialized. Unfortunately, I don't know enough of Haskell to pinpoint what parts of the code explain the difference between these cases ... 

**The context:** I have developped a Python library (https://github.com/boisgera/pandoc) that reads the pandoc-types data models (for as many versions of pandoc as possible) to reproduce automatically the equivalent hierarchy of classes in Python, so that json data can be exchanged with the available pandoc executable to work with a pandoc document representation in Python. The target being the people (first and foremost : me :wink: ) that need to analyze and transform a document with a nice AST and are fluent in Python but not so much in Haskell (or in Lua). To continue to do that, I need to be able to infer automatically from the output of `:browse Text.Pandoc.Definition` in `ghci` the JSON serialization rule for each data type. This is why a simple and mechanical rule would help !

Cheers,

SB

despresc:
No, there was no particular reason that I can remember. I think I just missed that `newtype`-defined types were serialized without their type information when I was looking at the other instance definitions.

despresc:
Or other single-constructor `data` types, for that matter.

jgm:
This will mean another version bump in pandoc-types, but I think it's probably worth making these changes so that we have a consistent JSON serialization scheme.

PR #76, opened 2020-08-30
Use untagged JSON encoding for single-constructor types

despresc:
This changes all of the single constructor `Table` types' `ToJSON` and `FromJSON` instances to the `Generic` ones, which has the effect of encoding them as arrays (for multi-argument constructors) or as the representation of the inner type (for `newtype` constructors).

All the other `newtype`-defined types in `Definition` have this encoding. The only other single constructor types in `Definition` are `Citation` (which is a record, and encoded as a JSON object) and `Pandoc` itself (which is encoded like a record, with additional metadata).

Closes: #75


despresc:
Well, the multi-argument constructors are the `Generic` instance. The `newtype` constructors have the `GeneralizedNewtypeDeriving` instance, but the effect is the same.

despresc:
Incidentally, `Meta` doesn't use the `newtype`-derived instances, and `Citation` doesn't use the `Generic` instances, when both give the same encoding as now. I'm not sure if you'd prefer explicit instances for the `Table` types, or if `Meta` and `Citation` should be changed to use the derived instances.

jgm:
Great. It would be worth adding some testEncodeDecode test cases (in test-pandoc-citeproc.hs) to cover these.
In the past we've had at least one case where a derived encoding changed due to changes in dependencies, and this would at least catch such thing.s

despresc:
That should be it. I've also found a small issue with the table model that is assumed in `Builder` that makes the table normalization in `Builder.table` a little too restrictive. I'll open a pull request soon, since that fix should probably be in the next version too (if you agree).

despresc:
I'm not sure if this is welcome, but the template haskell instances should be at least as efficient as what is there now. They are probably more efficient for `ToJSON`, since they automatically give the instances a definition of `toEncoding`. This should fix the `Map` determinism issues as well, if the `Attr` serialization format is to be changed from a list of pairs to an object.

Otherwise I can remove that most recent commit.

despresc:
The only thing that changed is that the keys for `Citation` now appear in the order in which they were defined, when serialized.

jgm:
I'm on board for using TH (simpler maintenance), I think.  I know that earlier (236e80120fb15d70584259843879e923743b3a54) we created the manual instances instead of deriving them, because changes in the dependencies affected the generated JSON.  It is important that the JSON format be stable and fixed.  But given the number of base versions we're testing with, this seems pretty solid.

jgm:
Note:  I've added this to master, bumped version to 1.22 (because of the JSON encoding changes, which could definitely be breaking for some users), and updated the changelog.  I'm thinking this should be released now, before we worry about the larger changes.  Does that seem right?

despresc:
That seems reasonable, though I'm not too familiar with the normal tempo of `pandoc-types` releases. It could be good to get that change out now, since it might take a while to discuss the other changes?

hadley:
This change seems to have had a fairly profound impact on the json AST, but I'm having a hard time figuring exactly what's changed because it's all explained in terms of Haskell types (which I'm not very familiar with). For my particular case, I _think_ it's that the children of `Table` no longer have `c` children, but their children are instead directly embedded? Does that sound plausible?

despresc:
That sounds right. All of the table components with no variants are now represented by whatever their `"c"` content was, since the `"t"` type tag is unnecessary for them. In particular, these components changed: the table caption, table head, table bodies, table foot, row head column number, column and row span numbers, rows, and cells.

The keys in citation objects also have a slightly different order now, which is less important.

PR #85, opened 2020-11-04
Fix incorrect table ColWidth documentation

pyssling:
The documentation stated that the ColWidth represented the width
of the column as a fraction of the table width when in represents
a percentage of the page width.

### DIFF
--- a/src/Text/Pandoc/Definition.hs
+++ b/src/Text/Pandoc/Definition.hs
@@ -224,8 +224,7 @@ data Alignment = AlignLeft
                | AlignCenter
                | AlignDefault deriving (Eq, Ord, Show, Read, Typeable, Data, Generic)
 
--- | The width of a table column, as a fraction of the total table
--- width.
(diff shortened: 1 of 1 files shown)


pyssling:
@despresc  Could you have a look at this? I see you are the author of the original code. This came up in a pandoc review - and checking the HTML Writer the intention is in fact that each ColWidth represents a percentage of the text-width - though that wasn't what the AST documented.

despresc:
I believe @jgm was the one who introduced `ColWidth` to the `Table` type. (I made unrelated changes to the `Table` type recently). For what it's worth, I think `pandoc` has interpreted it as being a fraction of the text width for some time now.

despresc:
Oh, sorry. I had forgotten that it was my commit that introduced the `ColWidth` data type. The column width stuff was still in `Table` before me. It's just that it was a `Double` originally.

jgm:
Thanks!  I guess the original documentation was due to me, but long ago.

PR #94, opened 2021-11-23
[API change] Add `Figure` block constructor

tarleb:
The new Figure block represents a figure with attributes, caption,
and arbitrary block content.

This combines code by @despresc and @argent0.

### DIFF
--- a/src/Text/Pandoc/Definition.hs
+++ b/src/Text/Pandoc/Definition.hs
@@ -254,7 +254,7 @@ data TableFoot = TableFoot Attr [Row]
 -- | A short caption, for use in, for instance, lists of figures.
 type ShortCaption = [Inline]
 
--- | The caption of a table, with an optional short caption.
+-- | The caption of a table or figure, with optional short caption.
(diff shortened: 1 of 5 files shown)


tarleb:
It's the structure assumed in https://github.com/jgm/pandoc/pull/7704.

src/Text/Pandoc/Definition.hs
```diff
@@ -301,6 +301,9 @@ data Block
     -- column alignments and widths (required), table head, table
     -- bodies, and table foot
     | Table Attr Caption [ColSpec] TableHead [TableBody] TableFoot
+    -- | Figure, with attributes, caption and caption position, width
+    -- (optional), and content (list of blocks)
+    | Figure Attr Caption [Block]
```
mb21, line 306: are `caption position and width (optional)` encoded in the `Attr` somehow?

jgm: I think this comment may be obsolete; an earlier version did represent caption position.

PR #95, opened 2022-03-08
Update ToJSONFilter docs: remove Meta, MetaValue

TravisCardwell:
Neither `Walkable Meta Pandoc` nor `Walkable MetaValue Pandoc` instances exist, so these types cannot be used with `ToJSONFilter`.

I grepped the source to check if there are any instances that I am not importing, and I did not find any.

I wrote a simple program to confirm that I am not missing something, and I indeed get errors for both types.

```haskell
{-# LANGUAGE TypeApplications #-}

module Main (main) where

import qualified Text.Pandoc.JSON as Pandoc

mainMeta :: IO ()
mainMeta = Pandoc.toJSONFilter $ id @Pandoc.Meta

mainMetaValue :: IO ()
mainMetaValue = Pandoc.toJSONFilter $ id @Pandoc.MetaValue

main :: IO ()
main = undefined
```

```
/path/to/Meta.hs:8:12: error:
    • No instance for (Text.Pandoc.Walk.Walkable
                         Pandoc.Meta Pandoc.Pandoc)
        arising from a use of ‘Pandoc.toJSONFilter’
    • In the expression: Pandoc.toJSONFilter $ id @Pandoc.Meta
      In an equation for ‘mainMeta’:
          mainMeta = Pandoc.toJSONFilter $ id @Pandoc.Meta
  |
8 | mainMeta = Pandoc.toJSONFilter $ id @Pandoc.Meta
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

/path/to/Meta.hs:11:17: error:
    • No instance for (Text.Pandoc.Walk.Walkable
                         Pandoc.MetaValue Pandoc.Pandoc)
        arising from a use of ‘Pandoc.toJSONFilter’
    • In the expression: Pandoc.toJSONFilter $ id @Pandoc.MetaValue
      In an equation for ‘mainMetaValue’:
          mainMetaValue = Pandoc.toJSONFilter $ id @Pandoc.MetaValue
   |
11 | mainMetaValue = Pandoc.toJSONFilter $ id @Pandoc.MetaValue
   |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

PR #96, opened 2022-03-10
Add Walkable Meta(Value) Pandoc instances

TravisCardwell:
I implemented the following instances, allowing `ToJSONFilter` to work with `Meta` and `MetaValue`, as suggested in #95.

* `Walkable Meta Pandoc`
* `Walkable MetaValue Pandoc`

Notes:

* The `Walkable Meta Pandoc` instance does not need to recurse because `Meta` is only used in `Pandoc`.  There is no `[Meta]` instance because there is only a single `Meta` value.
* The `Walkable MetaValue Pandoc` instance makes use of new `Walkable MetaValue Meta` and `Walkable MetaValue MetaValue` instances.  These instance implementations map over containers directly; there is no `[MetaValue]` instance because `MetaValue` is also used as `Meta` and `MetaMap` map values.

Please let me know if there are any issues or formatting inconsistencies that you would like me to correct.  You are of course welcome to fix such things yourself if doing so requires less of your time.

Thank you!

### DIFF
--- a/src/Text/Pandoc/JSON.hs
+++ b/src/Text/Pandoc/JSON.hs
@@ -86,7 +86,7 @@ import System.Environment (getArgs)
 -- to stdout.
 --
 -- For a straight transformation, use a function of type @a -> a@ or
--- @a -> IO a@ where @a@ = 'Block', 'Inline', or 'Pandoc'.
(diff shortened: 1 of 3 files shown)


jgm:
Looks good, thanks!

PR #99, opened 2022-05-19
Recreate the AST with functors

### LINKED ISSUE #98: "Functored" AST
Instead of doing e.g.
```haskell
data Block = ...Block...
```
what if we did 
```haskell
data BlockF block = ...block...
   deriving Functor
newtype Block = MkBlock (BlockF Block)
```

This trick allows ASTs to be extensible, among other benefits.

I generally like this approach, but I had a specific problem in mind that would benefit which I would like to share. https://github.com/obsidiansystems/dombuilder-pandoc/blob/master/src/Reflex/Dom/Builder/Pandoc.hs is some code to translate the Pandoc AST to reflex-dom "dom builder action" order to use within a website built with reflex dom.

I would like to have my own custom handling of pandoc the pandoc AST  --- e.g. parsing relative URLs in links into a Route AST --- without having to copy and paste that function. But if we do the above "functored AST" approach, I can replace
```haskell
block :: DomBuilder t m => Block -> m ()
```
with
```haskell
blockF :: DomBuilder t m => BlockF a -> m ()
```
or
```haskell
blockF :: DomBuilder t m => BlockF (m ()) -> m ()
```
These can be thought of as specialized versions of [`traverse_`](https://hackage.haskell.org/package/base-4.16.1.0/docs/Data-Foldable.html#v:traverse_) or [`sequence_`](https://hackage.haskell.org/package/base-4.16.1.0/docs/Data-Foldable.html#v:sequence_), and they are very nice to work with!

Ericson2314:
Fixes #98

See the issue for details of why this is useful


Ericson2314:
Oops, I had messed up the cabal file. Should be OK now.

jgm:
CI is failing for this and the associated PR on pandoc, so it's hard to evaluate.

Ericson2314:
OK will fix tests and CPP `COMPLETE` pragmas.

Ericson2314:
OK added a JSON deriving option so hopefully the newtypes don't introduced extra fluff. That was what was making the tests fail.

Ericson2314:
Oh good this does pass now.

Issue #104, opened 2022-11-28
Work with a StateT transformator


KommuSoft:
It would be nice if we can also use a `StateT` monad transformer when walking through the AST and thus keep track of the section, and update the state based on content that we see when "walking" over the tree.

jgm:
`walkM` in Text.Pandoc.Walk works for any monad.  So yes, you can use it with a state monad or something else with StateT.  Am I missing something in your question?

KommuSoft:
After rereading, I realized it was missing the most important part: the [`toJSONFilter`](https://hackage.haskell.org/package/pandoc-types-1.22.2.1/docs/Text-Pandoc-JSON.html#v:toJSONFilter) which does not seem to allow effectively walk over this.

The implementation looks like https://github.com/jgm/pandoc-types/blob/da9aeed5f3c5e7ad4d456e5b1e99e7a1573854be/src/Text/Pandoc/JSON.hs#L120-L124

so probably generalizing it to any monad transformer with `IO` as the "outer" one is possible. I will take a look into it.

jgm:
Yes, maybe we could generalize these to any MonadIO instance.
An alternative approach that should work already would be to use the IO Monad and an IORef for state.

jgm:
Closed by #105.

PR #105, opened 2022-11-29
work with any MonadIO fixing #104

KommuSoft:
Fixing #104

### DIFF
--- a/src/Text/Pandoc/JSON.hs
+++ b/src/Text/Pandoc/JSON.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE FlexibleInstances, FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances, FlexibleContexts, MultiParamTypeClasses #-}
 {-
 Copyright (c) 2013-2019, John MacFarlane
 
@@ -74,6 +74,7 @@ module Text.Pandoc.JSON ( module Text.Pandoc.Definition
 where
 import Text.Pandoc.Definition
 import Text.Pandoc.Walk
+import Control.Monad.IO.Class(MonadIO(liftIO))
 import Data.Maybe (listToMaybe)
 import qualified Data.ByteString.Lazy as BL
 import qualified Data.Text as T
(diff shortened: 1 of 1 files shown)


jgm:
@tarleb what do you think of this change?

One query.  Instead of
```hs
 class ToJSONFilter m a where
   toJSONFilter :: a -> m ()
```
would it make sense to do
```hs
class ToJSONFilter a where
  toJSONFilter :: forall m. MonadIO m => a -> m ()
```
?

src/Text/Pandoc/JSON.hs
```diff
@@ -99,32 +100,33 @@ import System.Environment (getArgs)
...
-class ToJSONFilter a where
-  toJSONFilter :: a -> IO ()
+class ToJSONFilter m a where
+  toJSONFilter :: a -> m ()
 
-instance (Walkable a Pandoc) => ToJSONFilter (a -> a) where
+instance (Walkable a Pandoc) => ToJSONFilter IO (a -> a) where
   toJSONFilter f = BL.getContents >>=
     BL.putStr . encode . (walk f :: Pandoc -> Pandoc) . either error id .
     eitherDecode'
```
jgm, line 110: Do we still need this?  After all, IO is an instance of MonadIO, so this is already covered by lines 111-115, isn't it?
Ditto for the other IO instance below.

tarleb:
I guess I'd prefer the `forall` version over the multiparam typeclass, but I don't feel strongly about it.

jgm:
Is there any particular reason favoring the forall approach?

tarleb:
I left out my reasoning intentionally, as I don't have a *good* answer for that. The universally quantified variant seems cleaner and more focused to me. The extra flexibility that comes with the multiparam typeclass variant probably won't be necessary for most users.

jgm:
I think most users won't touch this at all, or even know about it.

Seems to me that non-MonadIO instances are conceivable -- e.g. for Conduit or pipes or any streaming abstraction that can receive input and give output.

KommuSoft:
The original reason to do this was to also work for example with a `StateT IO Int Pandoc`, and thus can use this when working with the `toJsonFilter (walkM someFunc)` with `someFunc :: Pandoc -> StateT IO Int Pandoc` for example. We can work with `evalStateT 0 (walkM someFunc)` and use that with `toJsonFilter`, but that requires to initialize the state before the `walkM` and the state is "gone" after the `walkM` then. It would be nice if `toJsonFilter` could thus work with `StateT IO Int Pandoc` as well, to process this. The `forall` will likely not allow that, since it means it can *convert* it to *any* `MonadIO`.

jgm:
It should work fine either way with `StateT IO Int Pandoc` -- something that works for *all* MonadIO instances will work for this one.

KommuSoft:
Yes, but if `someFunc` thus is an `Pandoc -> StateT IO Int Pandoc`, then the `toJsonFilter` needs to be able to handle this, with the current implementation, it only handles a `Pandoc -> Pandoc` or a `Pandoc -> IO Pandoc`.

jgm:
Maybe I'm misunderstanding. If we have
```hs
class ToJSONFilter a where
  toJSONFilter :: forall m. MonadIO m => a -> m ()
```
Then `toJSONFilter` could indeed be applied to `someFunc` which has type `Pandoc -> StateT IO Int ()`, which is an instance of `forall m. Monadi IO m => a -> m()`.  It could *not* be applied to `Pandoc -> StateT IO Pandoc`, but neither could your interface
```hs
 class ToJSONFilter m a where
   toJSONFilter :: a -> m ()
```
because `()` doesn't match `Pandoc`.

jgm:
I'm going to merge this and remove the unnecessary IO instance in a second commit.

PR #107, opened 2023-01-10
CI: add GHC 9.4.4 with cabal 3.8 to test matrix, fix GHC 8.0 & 8.2.

tarleb:


jgm:
That's weird, why did we suddenly need libtinfo5?

tarleb:
Not sure, but I suspect the reason to be the changes to the ubuntu-latest runner that GitHub rolled out in autumn last year.

PR #108, opened 2023-01-13
Remove Null constructor from Block.

### LINKED ISSUE #91: Remove Null?
We have a Block constructor Null. I think this was added before we were using Builder in the readers; I wonder if there's any reason for it to exist.  Note that there's nothing like this for Inline.

tarleb:
[API change]

Closes #91.


tarleb:
This just reintroduces the commit that had been reverted.

jgm:
If we did this, would we want to filter out null elements from JSON arrays? (See discussion in #91)  That would mean not using the automatically generated instances for `[Block]`.

tarleb:
I think that might best be left to the filter library.

PR #110, opened 2023-01-22
Remove string qq dependency

rebeccaskinner:
This pull request replaces the dependency on the string-qq package with a compatible module. Depending on string-qq causes the transitive dependency graph to require `text >=1.2 && <1.3`. Removing this dependency will allow building with newer versions of `text` (tested with text-2.0.1). Given the very small size and limited use of `string-qq`, removing a dependency here seems like a win.

### DIFF
--- a/test/Data/String/QQ.hs
+++ b/test/Data/String/QQ.hs
@@ -1,0 +1,19 @@
+-- | This module is based off the QQ implementation from string-qq
+-- (https://github.com/audreyt/string-qq).
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE RankNTypes #-}
+module Data.String.QQ (s) where
+import Data.String (IsString(..))
+import Language.Haskell.TH.Quote (QuasiQuoter(..))
+
+s :: QuasiQuoter
+s = QuasiQuoter expr pat typ dec
+  where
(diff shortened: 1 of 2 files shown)


jgm:
This dependency only affects the test suite, so it shouldn't prevent programs that depend on pandoc-types from using text >= 2.  Still, since string-qq seems not to be currently maintained, I think this change is a good idea.

test/Data/String/QQ.hs
```diff
@@ -0,0 +1,19 @@
+-- | This module is based off the QQ implementation from string-qq
+-- (https://github.com/audreyt/string-qq).
```
jgm, line 2: What was the license for string-qq, and does this use conform to it?

rebeccaskinner: The original package is listed as public domain.

rebeccaskinner:
> This dependency only affects the test suite, so it shouldn't prevent programs that depend on pandoc-types from using text >= 2.

True. Sorry, I think my original motivation wasn't clear. I was trying to build the package with ghc 9.4 using nix, and found that the older version of text wasn't available in package set. It seemed like a win-win to push the fix here since it both makes the happy path for building with nix a bit easier and removed the unmaintained dependency.

jgm:
OK, thanks for this!

PR #115, opened 2023-11-15
Allow deepseq-1.5

Bodigrim:


jgm:
Thanks. I also made a Hackage revision.

PR #118, opened 2024-09-27
make ToJSONFilter instances that also pass the meta when walking the pandoc tree

KommuSoft:


jgm:
I'm not in favor of this, because as noted in your earlier query, you can simply use a Reader monad with the current API.

PR #119, opened 2025-12-29
Remove unused `ghc-prim` dependency

konsumlamm:


konsumlamm:
You might be right, although Hackage doesn't have such old versions of `ghc-prim`. It seems `GHC.Generics` was added in `base-4.6`, so `ghc-prim` might still be required for `base-4.5`. However, I suggest simply dropping support for `base-4.5`, it's ancient.

